PR #22, opened 2022-04-18
add more algos

montanalow:
This is most of the algos in scikit.

### DIFF
--- a/pgml/pgml/model.py
+++ b/pgml/pgml/model.py
@@ -1,8 +1,10 @@
 from re import M
 import plpy
-from sklearn.linear_model import LinearRegression, LogisticRegression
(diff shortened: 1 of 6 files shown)


pgml/pgml/model.py
```diff
@@ -463,9 +500,10 @@ def fit(self, snapshot: Snapshot):
         """
             )[0]
         )
+        self.__init__()
 
     def deploy(self):
-        """Promote this model to the active version for the project that will be used for predictions"""
+        """Promote t his model to the active version for the project that will be used for predictions"""
```
levkk, line 506:
```suggestion
        """Promote this model to the active version for the project that will be used for predictions"""

```

levkk (approved):

PR #28, opened 2022-04-18
Use the examples as the test suite

montanalow:
This also enables passing hyperparams to the algorithms to speed things up during tests.


levkk (approved):
Nice! Don't forget the red wine dataset references in the README :D

PR #35, opened 2022-04-22
admin works

montanalow:
Capture


CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+
+    $ pip freeze > requirements.txt
+
+7) Make sure requirements.txt has no changes, which indicates your virtual environment is setup correctly.
+
+    $ git status
+
+8) Create the development databases in the postgres cluster running inside docker:
+
+    $ psql -c "CREATE DATABASE pgml_development" postgres://postgres:postgres@127.0.0.1:5433/
```
levkk, line 31: We can do this in `bin/entrypoint.sh`

.env.TEMPLATE
```diff
@@ -0,0 +1,3 @@
+DJANGO_SECRET_KEY='insecure-development-key-for-django'
```
levkk, line 1: You can put these defaults directly into `settings.py`.

scripts/manage.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
```
levkk, line 1: This feels very Rails-like. The Django app should live in its own folder I think, I'm not familiar with this particular organization pattern in Python/Django. Maybe it would be best to place it in its own folder instead of the root of the repo? We will be storing a bunch more things in here as well, so we should organize it early.

For example, we can use the jungle book for inspiration (since we're all about elephants here!): https://lionking.fandom.com/wiki/Simba

Ideally, in the Python world, we want our packages to be as uniquely named as possible, so things like `control` or especially `www` in the global namespace make me think twice.

sql/install.sql
```diff
@@ -3,93 +3,6 @@ SET client_min_messages TO WARNING;
 -- Create the PL/Python3 extension.
 CREATE EXTENSION IF NOT EXISTS plpython3u;
 
----
```
levkk, line 6: This broke the Docker container from starting up correctly as the demo. This script installs everything into the `root` database, so maybe we can keep it around until we fully migrate to Django?

www/settings.py
```diff
@@ -0,0 +1,135 @@
...
+dotenv.load_dotenv()
+
+# Build paths inside the project like this: BASE_DIR / 'subdir'.
+BASE_DIR = Path(__file__).resolve().parent.parent
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/4.0/howto/deployment/checklist/
+
+SECRET_KEY = os.environ["DJANGO_SECRET_KEY"]
```
levkk, line 27: Put some defaults in here. We'll override them in prod for sure, but it'll help start it up locally for easy local development.

www/settings.py
```diff
@@ -0,0 +1,135 @@
...
+
+DEBUG = os.environ["DJANGO_DEBUG"] == "True"
+
+ALLOWED_HOSTS = []
```
levkk, line 31:
```suggestion
ALLOWED_HOSTS = [`*`]
```
This is not "safe", but it won't run in non-debug mode otherwise, at least not until we get a domain name to host it from.

www/settings.py
```diff
@@ -0,0 +1,135 @@
...
+
+PROJECT_PATH = os.path.realpath(os.path.dirname(__file__))
+
+WSGI_APPLICATION = 'www.wsgi.application'
+
+
+# Database
+# https://docs.djangoproject.com/en/4.0/ref/settings/#databases
+from urllib.parse import urlparse
+database = urlparse(os.environ["DATABASE_URL"])
```
levkk, line 82: Check out this extension: https://pypi.org/project/dj-database-url/

It also allows to specify a default for local dev.

control/static/css/base.css
```diff
@@ -0,0 +1,512 @@
...
+main * {
+    margin: 0 min(5vw, 10px);
+}
+
+/* Footer */
+footer {
+    height: 0;
+}
+
+@media only screen and (max-width: 690px) {
```
levkk, line 235: We might want to move away from these soonish and use something more mainstream, thought I'll admit I haven't touched this stuff in a long long time, hence my preference for libs I knew from back in the day that still exist, e.g. boostrap. We can't feasibly test every single browser and version, so we should delegate this responsibility to people who's full time job is to do that kind of stuff, e.g. bootstrap developers.

CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+    $ cp .env.TEMPLATE .env
+    $ nano .env
+
+10) Run db migrations
+
+    $ ./scripts/manage.py migrate
+
+11) Create the superuser
+
+    $ ./scripts/manage.py createsuperuser
```
levkk, line 47: We could include some fixtures for local dev, e.g. `admin@localdev` with password `password`. Will make setup quicker.

levkk:
The typical Django local experience should be:

```bash
python3 -m venv venv && source venv/bin/activate # Or whatever virtualenv you want to use
pip install -r requirements.txt
python manage.py migrate
python manage.py runsever
```

If we deviate from that, something smells imo. We already have default creds setup for our Postgres in Docker, maybe we can leverage that to startup faster?

levkk:
It works! 
image

levkk:
This feels like the step in the right direction. I think we need to tighten up the deployment a little bit, because this new experience will break the experience documented in the README, so we should maybe migrate everything to Django and present it in a docker-compose environment together with the `postgres` container.

control/views/deployments.py
```diff
@@ -0,0 +1,21 @@
+from django.shortcuts import render, get_object_or_404
+
+from .. import models
```
levkk, line 3: Much prefer absolute imports over relative ones, not sure why.

e.g.

```python
from control import models
```

CONTRIBUTING.md
montanalow: Cool. I think we also need to setup a script for non docker environments. I like developing outside of docker for iteration speed. Maybe we can have a single setup script that works for both.

CONTRIBUTING.md
montanalow: I'll setup the test suite to populate the development database by default.

CONTRIBUTING.md
levkk: Yeah for sure, local dev without docker is much better!

montanalow:
closing in favor of #38

PR #48, opened 2022-04-28
snapshot analysis rough draft

montanalow:
It's just a draft
Capture

### DIFF
--- a/pgml-admin/app/models.py
+++ b/pgml-admin/app/models.py
@@ -1,4 +1,4 @@
-from django.db import models
+from django.db import models, connection
 
 
 class Project(models.Model):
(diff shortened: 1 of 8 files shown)


pgml-admin/app/templates/snapshots/snapshot.html
```diff
@@ -11,13 +11,110 @@ <h1>{{ snapshot.relation_name }}</h1>
...
+    {% endfor %}
+  </tbody>
+</table>
+<script>
+
+{% for name, feature in features.items %}
+  var {{name}}_samples = {{feature.samples}};
+{% endfor %}
+var layout = {};
+function linearRegression(x,y){
```
montanalow, line 60: I need to move this into postgres.

levkk: Oh interesting!

pgml-extension/pgml_extension/model.py
```diff
@@ -204,9 +204,11 @@ def create(
         plpy.execute(
             f"""
             CREATE TABLE pgml."snapshot_{snapshot.id}" AS 
-            SELECT * FROM {snapshot.relation_name};
+            SELECT * FROM {snapshot.relation_name}
+            ORDER BY random();
         """
```
levkk, line 209: Why does everyone order by random, curious what that does?

pgml-admin/app/models.py
```diff
@@ -21,13 +20,32 @@ class Snapshot(models.Model):
...
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
     class Meta:
         db_table = '"pgml"."snapshots"'
         managed = False
 
+    def sample(self, limit=1000):
+        with connection.cursor() as cursor:
+            cursor.execute(f"SELECT * FROM {self.relation_name} LIMIT %s", [limit])
```
levkk, line 34: This will give you generally the same set when called multiple times due to caching/rows in shared buffers. If you want more random samples, not sure what the best way would be to go about it.

levkk (approved):
Nice!

levkk:
Oh hey, our CI works! :)

pgml-admin/app/models.py
montanalow: That's what the order by random is for initially. It's OK, even desireable-ish that it's the same set on follow up reloads.

PR #78, opened 2022-05-04
A few fixes and adding DRF

levkk:
1. Fix migrations by adding `public` to search path for Django.
2. Remove unused property `y_column_type` from `Snapshot`.
3. Added Django Rest Framework under `/api` in read-only mode. This helps if we want to fetch things dynamically via XHR, and also to understand what data we have in the database. It's available at https://localhost:8000/api.


pgml-admin/pgml_admin/settings.py
```diff
@@ -91,7 +92,7 @@
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql",
-        "OPTIONS": {"options": "-c search_path=pgml"},
```
montanalow, line 94: This is something I worry slightly about. I'd rather be explicit in every query that reaches out to the host schema's from admin. For example, imagine we use the django sessions backed by the database, and the host db has their own db backed sessions table. If they forget to run migrations, we may read the wrong sessions table in admin. I would rather fail with a table not found error.

montanalow (commented):
I'm curious what breakage you saw in migrations without the public path in the schema.

pgml-admin/pgml_admin/settings.py
levkk: Good point. I moved the Django tables into the `pgml` schema:

```
pgml_development=# select count(1) from pgml.django_migrations;
 count 
-------
    19
(1 row)
```

montanalow (approved):

PR #79, opened 2022-05-04
add basic vector operations

montanalow:


levkk (approved):
Very cool. I'm going to start thinking about how we can add some tests. Postgres extensions and Postgres itself has "output tests" where they just run some test SQL and compare the output to what they expect, we could try that eventually.

PR #98, opened 2022-05-07
Cleanup blame view in GitHub.

samdobson:
This prevents #97 from obscuring change history (ref:  [GitHub docs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view))


montanalow:
Thanks!

PR #99, opened 2022-05-07
Use consistent values for `test_size`

samdobson:
This PR sets the default train-test split to 0.25, to match the [scikit-learn default](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html), and to conform to the behaviour described in the [snapshot docstring](https://github.com/postgresml/postgresml/blob/0668cd338bac34569229b6913259bc8d444fe158/pgml-extension/pgml_extension/model.py#L179).


montanalow:
Thanks!

PR #100, opened 2022-05-07
Additional classification metrics.

samdobson:
This PR calculates a few additional metrics that are useful for evaluating the performance of classification models, namely:

* Accuracy
* ROC AUC
* Log loss

### DIFF
--- a/pgml-extension/pgml_extension/model.py
+++ b/pgml-extension/pgml_extension/model.py
@@ -17,6 +17,9 @@ from sklearn.metrics import (
     f1_score,
     precision_score,
     recall_score,
+    roc_auc_score,
+    accuracy_score,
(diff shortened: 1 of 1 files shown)


montanalow:
Thanks! Medium term we'd like to expose all of the [scoring metrics](https://scikit-learn.org/stable/modules/model_evaluation.html#scoring-parameter) and custom key_metrics as training options, but this is a healthy step in that direction. In addition we can also add some graphs to the admin UI for these metrics.

samdobson:
Nice!

In the early stages of evaluating a project I tend to make small PRs that scratch my own itches, but I'm open to a more substantial collaboration if we can align on goals (happy to chat).

PR #102, opened 2022-05-09
New project wizard

levkk:
Adding basic new project wizard.

https://user-images.githubusercontent.com/9561483/167773799-e102a9de-d036-4a7f-9c75-04fa30541f87.mp4

### Design
The algorithm selection is using buttons because I expect us to add hyperparameters to each of them, and I think it would be easier to add details into those than if we used a checkbox list.

### TODOs
1. Add a data explorer & SQL editor to be able to create views without connecting with `psql`.
2. Create a snapshot prior to training and show the snapshot analysis before picking algorithms.
3. Replace the cheesy loader with a progress bar indicating which algorithm is currently being trained.
4. For large datasets or high latency data sources, create a snapshot creation progress indicator.
5. Graceful error handling.

### DIFF
--- a/pgml-admin/app/static/css/base.css
+++ b/pgml-admin/app/static/css/base.css
@@ -57,7 +57,7 @@ footer, header, hgroup, menu, nav, section {
 }
 body {
 	line-height: 1;
-    height: 100vh;
+    min-height: 100vh;
 }
 ol, ul {
 	list-style: none;
--- a/pgml-admin/app/static/css/base.css
+++ b/pgml-admin/app/static/css/base.css
@@ -57,6 +57,7 @@ footer, header, hgroup, menu, nav, section {
 }
 body {
 	line-height: 1;
+    height: 100vh;
 }
 ol, ul {
 	list-style: none;
--- a/pgml-admin/app/static/css/new.css
+++ b/pgml-admin/app/static/css/new.css
@@ -59,12 +59,15 @@ button, a[role=button] {
     color: initial;
 }
 
-button:hover, a[role="button"]:hover {
+button, a[role="button"] {
     cursor: pointer;
+}
+
+button:hover, a[role="button"]:hover {
     background-color: var(--gray-7);
 }
 
-button[disabled]:hover, a[disabled]:hover {
+button[disabled], button:disabled, a[disabled], a:disabled {
     cursor: not-allowed;
 }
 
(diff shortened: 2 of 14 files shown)


pgml-admin/app/static/css/base.css
```diff
@@ -57,6 +57,7 @@ footer, header, hgroup, menu, nav, section {
 }
 body {
 	line-height: 1;
+    height: 100vh;
```
montanalow, line 60: I don't know if this is safe, how does it work with overflow?

pgml-admin/app/templates/projects/index.html
```diff
@@ -14,5 +14,9 @@ <h1><span class="material-symbols-outlined">apps</span>Projects</h1>
     </li>
     {% endfor %}
   </ol>
+
+  <div class="button-container">
```
montanalow, line 18: Why do you need the div?

montanalow (commented):
At first glance this all looks pretty reasonable to me. Can't wait to see it in action.

pgml-admin/app/static/css/new.css
```diff
@@ -0,0 +1,77 @@
+input {
+    padding: 10px;
+    font-size: 1em;
+    outline: none;
+    border: 1px solid var(--gray-8);
+    border-radius: 1px;
+}
+
+.progress-bar {
```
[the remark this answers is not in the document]
levkk: Oh cool, didn't know this existed.

pgml-admin/app/static/css/base.css
levkk: Right, good call, will replace with `min-height`

pgml-admin/app/templates/projects/index.html
levkk: For the flex box

pgml-admin/app/views/projects.py
```diff
@@ -60,3 +67,208 @@ def get_context_data(self, **kwargs):
...
+
+    @action(detail=False, permission_classes=[], methods=["POST"])
+    def train(self, request):
+        """Train a new project."""
+        serializer = NewProjectSerializer(data=request.data)
+        if serializer.is_valid():
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    """
+                    SELECT * FROM pgml.train_joint(
```
montanalow, line 82: I'd use `train` here, unless you're letting them add multiple targets in the UX? I didn't see multiselect at that step. I'm not crazy about the current train/train_joint API split. It only exists to mirror `predict_joint`, because we can't have different return types (REAL vs REAL[]) on `predict()` and `predict_joint()`.

pgml-extension/pgml_extension/model.py
```diff
@@ -753,6 +753,18 @@ def predict(self, data: list):
         return y[0]
 
 
+def snapshot(
```
montanalow, line 756: Why define this at the root level, when it's just a pass through to the class?

montanalow (approved):
🎉

pgml-extension/pgml_extension/model.py
levkk: Excellent question...

pgml-admin/app/views/projects.py
levkk: Yup multi-select is allowed in the UI.

pgml-extension/pgml_extension/model.py
levkk: Will clean up later. I'm still thinking about this one.

PR #103, opened 2022-05-09
Repeatable model training

samdobson:
Goal of this PR is to be able to fit a classifier twice, to the same training data, and get the same result.

This code currently does not work, and it's not immediately clear why.
_Edit: now fixed (see below)_

### DIFF
--- a/pgml-extension/pgml_extension/model.py
+++ b/pgml-extension/pgml_extension/model.py
@@ -803,6 +803,8 @@ def train(
     # Model
     if algorithm_name is None:
         algorithm_name = "linear"
+    if not hyperparams.get("random_state"):
+        hyperparams["random_state"] = 0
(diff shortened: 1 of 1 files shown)


montanalow:
[random ordering is here](https://github.com/postgresml/postgresml/blob/master/pgml-extension/pgml_extension/model.py#L228)

samdobson:
Thanks both. By not passing the relation second time around, this is now working as expected and model training is now reproducible by default.

levkk:
Ah so `random_state` is not an argument accepted by all algorithms it seems.

samdobson:
My bad, will fix...

montanalow:
I found the issue while I was fixing test_sampling so included the additional fix in #104.

samdobson:
Apologies and thanks. Note to self: run tests first!

PR #118, opened 2022-05-13
Migrate to Debian

levkk:
1. Migrate to debian images
2. Rebuild the world on each `docker-compose up`, remove the volume.

### DIFF
--- a/pgml-extension/docker/entrypoint.sh
+++ b/pgml-extension/docker/entrypoint.sh
@@ -20,10 +20,7 @@ echo "Creating user and database..."
 psql -d pgml_development -f sql/install.sql -U postgres -h 127.0.0.1 > /dev/null
 
 echo "Installing example datasets"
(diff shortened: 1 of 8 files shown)


montanalow:
Works on ARM!

pgml-extension/docker/entrypoint.sh
```diff
@@ -18,6 +19,9 @@ echo "Creating user and database..."
 (createdb -U postgres -h 127.0.0.1 pgml_development 2> /dev/null) || true
 psql -d pgml_development -f sql/install.sql -U postgres -h 127.0.0.1 > /dev/null
 
+echo "Installing example datasets"
+psql -d pgml_development -f examples/classification.sql -U postgres -h 127.0.0.1 -f sql/test.sql -P pager
```
montanalow, line 23: Why not bring it all in for show and tell?

levkk: That runs everything.

PR #119, opened 2022-05-13
Quick prediction

levkk:
1. Quick prediction module in Model page.
2. `pgml.model_predict(model_id BIGINT, features DOUBLE PRECISION[])` for predicting against a specific model for debugging.

https://user-images.githubusercontent.com/9561483/168343255-2313d545-fb15-44e8-bc6f-65573f574df0.mp4

### DIFF
--- a/pgml-dashboard/app/views/models.py
+++ b/pgml-dashboard/app/views/models.py
@@ -2,11 +2,17 @@ from typing import OrderedDict
 from collections import namedtuple
 import json
 import logging
+import pickle
+import io
 
 from django.shortcuts import render, get_object_or_404
 from django.views.generic import DetailView, ListView
 from django.utils.safestring import SafeString
+from django.db import connection
+
 from rest_framework import viewsets
+from rest_framework.decorators import action
+from rest_framework.response import Response
 
 from app.models import Model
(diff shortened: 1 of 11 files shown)


pgml-dashboard/app/views/models.py
```diff
@@ -2,11 +2,17 @@
 from collections import namedtuple
 import json
 import logging
+import pickle
```
montanalow, line 5: is this actually used?

montanalow (approved):

montanalow:
This reminds me, we need to add Shap to the backlog! https://github.com/slundberg/shap

pgml-dashboard/app/views/models.py
levkk: No, good call

levkk, line 5:
```suggestion
```

pgml-dashboard/app/views/models.py
```diff
@@ -2,11 +2,16 @@
 from collections import namedtuple
 import json
 import logging
+import io
```
levkk, line 5:
```suggestion
```

PR #120, opened 2022-05-13
Base image for pgml-extension

levkk:
Faster startup time for `pgml-extension` by using a pre-built Docker image.


levkk:
Does not support ARM, closing.

PR #131, opened 2022-05-15
Add debounce to new project wizard; fix quick predictions

levkk:
1. Add `_.debounce` to fetch calls in new project wizard.
2. Add support in Quick Prediction for models with multiple labels.

### DIFF
--- a/pgml-extension/sql/install/models.sql
+++ b/pgml-extension/sql/install/models.sql
@@ -200,9 +200,16 @@ CREATE OR REPLACE FUNCTION pgml.model_predict(
 	model_id BIGINT, -- `id` from `pgml.models`
 	features DOUBLE PRECISION[] -- list of features that the model accepts
 )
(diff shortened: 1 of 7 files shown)


pgml-extension/sql/install/models.sql
```diff
@@ -200,9 +200,16 @@ CREATE OR REPLACE FUNCTION pgml.model_predict(
 	model_id BIGINT, -- `id` from `pgml.models`
 	features DOUBLE PRECISION[] -- list of features that the model accepts
 )
-RETURNS DOUBLE PRECISION
+RETURNS DOUBLE PRECISION[]
```
montanalow, line 203: This is the only real different with predict_multi. Why not use that instead?

levkk: `predict_multi` is not model specific, this one is.

montanalow (approved):

PR #147, opened 2022-05-17
binary classification example

montanalow:


montanalow:
@tylerdavis I added a test case for a binary classification example, since it seems to be what you were attempting that was catching issues.

PR #153, opened 2022-05-30
Hugging Face Integration

montanalow:
Harmonize existing models with hugging face to avoid multiple ways of doing things across different model providers
- rename objective to task 
- move serialized artifacts out of the database into files


pgml-extension/pgml_extension/utils.py
```diff
@@ -0,0 +1,47 @@
...
+        time = datetime.now() - start
+
+        _nested_timers -= 1
+        if _nested_timers == 0:
+            _ascii_pipes = ''
+        else:
+            delta = (_nested_timers - _previous_timer_level)
+            length = _nested_timers * 2
+            if delta < 0:
+                _ascii_pipes = _ascii_pipes[0:length]
```
levkk, line 35: Careful with these, they don't render well in logging systems that don't have a normal terminal, e.g. Datadog logs, and they make the logs hard to read.

pgml-extension/pgml_extension/model.py
```diff
@@ -640,16 +741,25 @@ def snapshot(self):
             self._snapshot = Snapshot.find(self.project_id)
         return self._snapshot
 
+    @property
+    def path(self):
+        return os.path.join(".", "postgresml", "models", str(self.id))
```
levkk, line 746: This does not work with replicas.

pgml-extension/pgml_extension/model.py
```diff
@@ -729,15 +816,43 @@ def fit(self, snapshot: Snapshot):
                     roc_auc_y_prob = y_prob[:, 1]
                 metrics["roc_auc"] = roc_auc_score(y_test, roc_auc_y_prob, average="weighted", multi_class="ovo")
 
-        # Save the model
+        self.metrics = metrics        
+
+        # Save the results
+        if not os.path.exists(self.path):
+            os.makedirs(self.path)
+        with open(self.pickle_path, "wb") as f:
```
levkk, line 824: Not ACID, can silently break because of lack of fsync call, cannot be replicated to replicas.

pgml-extension/sql/install/schema.sql
```diff
@@ -116,6 +115,38 @@ CREATE INDEX IF NOT EXISTS deployments_project_id_created_at_idx ON pgml.deploym
...
+---
+--- Distribute serialized models consistently for HA
+---
+CREATE TABLE IF NOT EXISTS pgml.files(
+	id BIGSERIAL PRIMARY KEY,
+	model_id BIGINT NOT NULL,
+	path TEXT NOT NULL,
+	created_at TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT clock_timestamp(),
+	updated_at TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT clock_timestamp(),
+	part_0 BYTEA NOT NULL,
```
levkk, line 127: Could turn this into rows instead, gives us up to 32TB of model storage.

pgml-extension/pgml_extension/utils.py
montanalow: We should probably get smart about logging depending on the environment.

PR #166, opened 2022-06-15
Fixed the link to the quickstart

morkovka3:
If you go to the main page of the website and click on "Get Started", it would show 404 Page Not Found. I just changed the endpoint to make it point to the right page.


montanalow:
Thanks for the fix!

PR #188, opened 2022-08-09
fraud example

montanalow:
Need to port into notebooks when ready


pgml-docs/docs/gym/fraud.md
```diff
@@ -0,0 +1,460 @@
...
+  perishable BOOLEAN
+);
+```
+
+```sql
+INSERT INTO PRODUCTS (name, price, perishable) 
+VALUES
+  ('1oz gold bar', '$1999.99', false),
+  ('a tale of 2 cities', '$19.99', false),
+  ('head of lettuce', '$1.99', true);
```
levkk, line 27: Use `RETURNING *` so the user can see the result of this query without having to query the table again.

pgml-docs/docs/gym/fraud.md
```diff
@@ -0,0 +1,460 @@
...
+  task => 'classification', -- we want to classify into true or false
+  relation_name => 'fraud_samples', -- our view of the data
+  y_column_name => 'fraudulent' -- the "labels"
+);
+```
+
+Oops. We're going to get an error:
+
+```
+ERROR:  ValueError: This solver needs samples of at least 2 classes in the data, but the data contains only one class: False
```
levkk, line 202: image

Got that instead.

levkk: image

levkk (approved):

PR #203, opened 2022-08-12
bust the project cache on a deploy

montanalow:
lots of extraneous changes relating to black

### DIFF
--- a/pgml-extension/pgml_extension/model.py
+++ b/pgml-extension/pgml_extension/model.py
@@ -205,7 +205,8 @@ class Project(object):
         global _project_cache, _last_deploy_id
 
         if last_deploy_id is not None and _last_deploy_id != last_deploy_id:
-            _project_cache = {}
+            for project in _project_cache.values():
+                project.expire_cached_deployed_model()
             _last_deploy_id = last_deploy_id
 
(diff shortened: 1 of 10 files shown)


pgml-extension/pgml_extension/model.py
```diff
@@ -916,10 +935,43 @@ def fit(self, snapshot: Snapshot):
...
+            f"""
+        CREATE OR REPLACE FUNCTION pgml.predict(
+            project_name TEXT,          -- Human-friendly project name
+            features DOUBLE PRECISION[] -- Must match the training data column order
+        )
+        RETURNS DOUBLE PRECISION
+        AS $$
+            from pgml_extension.model import Project
+
+            return float(Project.find_by_name(project_name, {q(deployment_id)}).deployed_model.predict(features))
```
levkk, line 958: Isn't deployment id model-specific? In this case, we are treating the `pgml.predict` function as model-specific but it's not.

montanalow: The original implementation uncached all projects, when any model was deployed so it was overly aggressive. Now we'll check if the project's model id has changed after a deploy, and only reload the model if it is no longer the currently deployed one.

PR #208, opened 2022-08-16
add a form select for cell type

montanalow:


montanalow:
TODO: refactor the controller to be specific to each cell, rather than the entire notebook, so we don't need to do DOM queries to find related elements.

pgml-dashboard/notebooks/static/notebooks/js/notebook.js
```diff
@@ -227,7 +230,6 @@ export default class extends Controller {
     const entries = []
 
     for (let entry of form.entries()) {
-      if (entry[0] === 'csrfmiddlewaretoken')
```
levkk, line 230: Make sure not to include the textarea input by accident, it's empty, the contents come from CodeMirror (see code below)...

levkk: If you want to use the form as-is, issue `this.newCellCodeMirror.save()` before collecting data from the form.

montanalow: Is that why you're not using a normal hotwire form submit/render cycle?

pgml-dashboard/notebooks/static/notebooks/js/notebook.js
```diff
@@ -71,10 +66,18 @@ export default class extends Controller {
     this.notebookNameTarget.classList.add('hidden')
   }
 
-  detectSql(codeMirror) {
```
levkk, line 74: You need to run this when the cell is edited because it defaults to `sql` even if it's markdown, enabling wrong syntax highlighting.

pgml-dashboard/notebooks/static/notebooks/js/notebook.js
levkk: That and also hotwire re-renders the whole page which resets the scrolling position making it hard to work with the notebook with multiple cells.

montanalow: I think that can be fixed with individual controllers per cell rather than one controller for the full notebook

PR #210, opened 2022-08-17
Play button

levkk:
- Introducing Turbo frames - things are so much simpler now.
- Play button to execute cell contents at will.
- Reset button to reset all cells to empty (not executed) state.
- A feel-good refactor.


pgml-dashboard/notebooks/templates/notebooks/cell.html
```diff
@@ -1,61 +1,82 @@
...
+        <div class="notebook-button">
+          <form action="{% url 'notebooks/cell/remove' notebook.pk cell.pk %}" method="post">
+            {% csrf_token %}
+            <button type="submit" title="Delete cell">
+              <span class="material-symbols-outlined">delete_forever</span>
+            </button>
+          </form>
+        </div>
+        <div class="notebook-button">
+          <form action="{% url 'notebooks/cell/edit' notebook.pk cell.pk %}?bust_cache={{ bust_cache }}" method="get">
```
levkk, line 66: If you have any ideas how to get around this using a cleaner method, that would be great. The problem here is when we submit the `GET` form, Turbo annotates the frame as coming from `notebooks/<pk>/cell/<pk>/edit` URL  which won't change. When we edit the cell again, Turbo doesn't execute the fetch because it thinks the result will be the same, so editing doesn't work twice on the same cell. Adding a unique parameter to the URL "busts that cache": Turbo can't make any assumptions about the content and executes the fetch.

PR #239, opened 2022-08-22
Typo in welcome tutorial

levkk:


montanalow (commented):
I think we must be using different versions of pyyaml or something else that changes "\n" from the literal to the string version, creating formatting differences.

levkk:
> I think we must be using different versions of pyyaml or something else that changes "\n" from the literal to the string version, creating formatting differences.

WSL vs Linux?

PR #240, opened 2022-08-22
Selected cell

levkk:
image

image

`Shift-Enter` to execute selected cell.

### DIFF
--- a/pgml-dashboard/app/static/css/base.css
+++ b/pgml-dashboard/app/static/css/base.css
@@ -602,6 +602,10 @@ main turbo-frame:first-of-type .notebook-cell {
   margin-bottom: 2px;
 }
 
+.notebook-cell.selected {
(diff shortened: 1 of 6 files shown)


pgml-dashboard/app/static/css/base.css
```diff
@@ -602,6 +602,10 @@ main turbo-frame:first-of-type .notebook-cell {
   margin-bottom: 2px;
 }
 
+.notebook-cell.selected {
+    border: 2px solid green;
```
montanalow, line 606: Not a fan of plain green, but we can leave it for the designer.

A very light gray shading on cell hover background might also help to scope the button actions visually.

levkk: We need to think of accessibility. Light colors are barely visible.

PR #247, opened 2022-08-23
Consistency for notebooks

montanalow:
Use consistent urls, templates, javascript controllers and other naming conventions, to make relative urls etc less thought provoking.

And it saves 10 lines of code overall.


levkk:
I put the notebooks into their own app on purpose.

PR #271, opened 2022-08-24
blog

montanalow:


pgml-docs/docs/blog/benchmarking.md
```diff
@@ -0,0 +1,207 @@
...
+		RETURN SQRT(SUM(squared.values))
+		FROM (SELECT UNNEST(vector) * UNNEST(vector) AS values) AS squared;
+	END
+$$;
+```
+
+```sql
+SELECT plpgsql_norm_l2(vector) FROM embeddings;
+```
+120 ms
```
levkk, line 91: Try switching to sql/output formatting we use elsewhere and maybe have the output of \timing there as well? In bold or something. Could look good.

PR #272, opened 2022-08-25
Reordering fixes

levkk:
- Reorder cells after delete (although the rendering doesn't work yet - have to refresh page)
- Delete cell is now double confirmation instead of automatic 10 second timer.
- Fix undo to be consistent with other cell rendering. Added a red left border.

### DIFF
--- a/pgml-dashboard/app/templates/notebooks/cell.html
+++ b/pgml-dashboard/app/templates/notebooks/cell.html
@@ -1,4 +1,6 @@
-<turbo-frame id="cell_{{ cell.cell_number }}">
+<turbo-frame id="cell_{{ cell.pk }}">
+  {% if cell.deleted_at %}
+  {% else %}
   <section data-controller="notebook-cell" class="notebook-cell {{ cell.get_cell_type_display|lower }}{% if selected %} selected{% endif %}" data-action="click->notebook#selectCell" data-notebook-target="cell">
 
     {% if edit %}
@@ -87,4 +89,5 @@
       {% endif %}
     </div>
   </section>
+  {% endif %}
 </turbo-frame>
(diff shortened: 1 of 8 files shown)


pgml-dashboard/app/templates/notebooks/cell.html
```diff
@@ -1,4 +1,6 @@
-<turbo-frame id="cell_{{ cell.cell_number }}">
+<turbo-frame id="cell_{{ cell.pk }}">
+  {% if cell.deleted_at %}
```
montanalow, line 2: If you're filtering these at the query level why do you need this additional check?

pgml-dashboard/app/templates/notebooks/cell.html
```diff
@@ -1,4 +1,6 @@
-<turbo-frame id="cell_{{ cell.cell_number }}">
+<turbo-frame id="cell_{{ cell.pk }}">
```
montanalow, line 1: ah yeah, we should always use pk everywhere rather than mutable things

pgml-dashboard/app/templates/notebooks/cell.html
montanalow: Also, is there some sort of extension that adds that filter by default everywhere?

montanalow (approved):

pgml-dashboard/app/templates/notebooks/cell.html
levkk: Because when we delete a frame, Turbo complains that the frame is gone when we re-render the page. TLDR: I haven't figured out how to delete frames so I'm returning an empty frame when deleting it. It's temporary, the frame is gone on page reload / new page load.

PR #301, opened 2022-09-04
start integrating smartcore for common algos in rust

montanalow:


pgml-extension/pgml_rust/src/lib.rs
```diff
@@ -159,88 +165,169 @@ mod pgml_rust {
...
+            } else if task == ProjectTask::classification {
+                let estimator = smartcore::linear::logistic_regression::LogisticRegression::fit(
+                    &x_train,
+                    &y_train,
+                    Default::default(),
+                )
+                .unwrap();
+                save(estimator, x_test, y_test, algorithm, project_id)
+            } else {
+                0
```
levkk, line 280: `error!` is best here, we don't want hidden errors that are hard to debug.

levkk:
Machine!!!

PR #332, opened 2022-09-20
checkpoint

montanalow:


pgml-extension/pgml_rust/src/engines/smartcore/mod.rs
```diff
@@ -0,0 +1,38 @@
+pub mod linear;
+
+use crate::orm::algorithm::Algorithm;
+use crate::orm::dataset::Dataset;
+use crate::orm::estimator::Estimator;
+use crate::orm::task::Task;
+use ndarray::{Array1, Array2};
+
+fn smartcore_train(
```
levkk, line 9: This is gone now, it's in `smartcore.rs`.

PR #333, opened 2022-09-20
Scikit hyperparameter search

levkk:


pgml-extension/pgml_rust/sql/schema.sql
```diff
@@ -81,7 +81,7 @@ CREATE TABLE IF NOT EXISTS pgml_rust.models(
 	hyperparams JSONB NOT NULL,
 	status TEXT NOT NULL,
 	metrics JSONB,
-	search pgml_rust.search,
```
montanalow, line 84: This should stay as an enum value, you can convert to_string() in the call to scikit.

levkk: That doesn't work, it gives me an error about not finding the type. I don't know why but enums are a pain.

montanalow: The other enums work, you just have to cast to TEXT on the in and out. We don't have the OID for them AFAIK, although pgx might know better.

levkk: It's probably the nullable property that's causing this then.

PR #357, opened 2022-10-04
Full LightGBM support in 2.0

montanalow:


pgml-extension/pgml_rust/src/bindings/lightgbm.rs
```diff
@@ -82,24 +81,27 @@ impl Bindings for Estimator {
...
-        //         Task::regression => results[0] as f32,
-        //     }
+        let mut results: Vec<f32> = results.into_iter().map(|i| i as f32).collect();
+        
+        let num_class = self.estimator.num_class().unwrap() as usize;
+        if num_class > 2 {
+            let mut argmax_results = Vec::with_capacity(results.len() / num_class);             
+            let mut max_i: usize = 0;
+            let mut max_probability = 0.0;
+            for (i, &probability) in results.iter().enumerate() {
```
levkk, line 91: Use `.chunks`: https://doc.rust-lang.org/stable/std/slice/struct.Chunks.html

PR #358, opened 2022-10-04
serialize num_classes for predictions

montanalow:


pgml-extension/pgml_rust/src/bindings/lightgbm.rs
```diff
@@ -123,13 +133,15 @@ impl Bindings for Estimator {
         Self: Sized,
     {
         let num_features = u64::from_be_bytes(bytes[..8].try_into().unwrap()) as usize;
+        let num_classes = u64::from_be_bytes(bytes[8..16].try_into().unwrap()) as usize;
```
levkk, line 136: We store this in our `pgml.models`, we can just pass that in at creation time by modifying the interface, and it'll be protected by the WAL and the type system.

PR #386, opened 2022-10-08
Enable Scikit by default but make it optional

levkk:
Some older systems may not have the Python version we want (Python 3.7+). We should not deprive them of the ability to run this extension, they will still benefit from XGBoost, LightGBM and Linfa.

### DIFF
--- a/pgml-extension/Cargo.toml
+++ b/pgml-extension/Cargo.toml
@@ -7,14 +7,14 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [features]
-default = ["pg13", "sklearn"]
+default = ["pg13", "python"]
 pg10 = ["pgx/pg10", "pgx-tests/pg10" ]
 pg11 = ["pgx/pg11", "pgx-tests/pg11" ]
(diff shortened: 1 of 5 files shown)


montanalow:
Do we need the same for huggingface?

pgml-extension/Cargo.toml
```diff
@@ -7,13 +7,14 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [features]
-default = ["pg13"]
```
montanalow, line 10: Since this is actually python (including huggingface)? I'd rename this feature to `python`

levkk: Yeah true that.

pgml-extension/src/orm/model.rs
```diff
@@ -154,6 +154,13 @@ impl Model {
                     _ => todo!(),
                 },
             },
+
+            #[cfg(not(feature = "sklearn"))]
+            Runtime::python => {
+                error!("Python runtime not supported, recompile with `--features scikit`")
```
montanalow, line 160: typo in scikit that you'll fix in the rename to `python`

PR #394, opened 2022-10-11
add wine and linnerud datasets

montanalow:
This also moves the toy datasets into pgml proper, without needing any external dependencies.


pgml-extension/src/orm/dataset.rs
```diff
@@ -380,133 +170,452 @@ pub fn load_breast_cancer(limit: Option<usize>) -> (String, i64) {
...
+        target FLOAT4
+    )",
+    );
+
+    let limit = match limit {
+        Some(limit) => limit,
+        None => usize::MAX,
+    };
+
+    let data: &[u8] = std::include_bytes!("datasets/diabetes.csv.gz");
```
levkk, line 328: This is loaded into the `.so`, so something to think about. Our `.so` is already pretty large (~18MB), I am not sure how that impacts performance.

montanalow:
I checked gzip size, combined, these are ~100k. I agree though, for the
larger datasets we’ll want to download instead.

PR #396, opened 2022-10-11
complete v2 end-to-end suite

montanalow:


pgml-extension/docker/entrypoint.sh
```diff
@@ -15,6 +15,8 @@ echo "Creating user and database..."
 (createdb -U postgres -h 127.0.0.1 pgml_development 2> /dev/null) || true
 
 echo "Installing pgml extension..."
+psql -U postgres -h 127.0.0.1 pgml_development -c 'DROP EXTENSION IF EXISTS pgml'
```
levkk, line 18: This is used by the quick start Docker image, so if you keep that there, it'll reset the whole thing every reboot, which isn't a terrible thing either.

PR #419, opened 2022-10-15
Limit XGBoost inference concurrency

levkk:
1. Add `update_params` method to `Bindings` which will update hyperparameters on loaded estimator. This is currently used for controlling XGBoost's concurrency (`nthread`).
2. Add `clear_cached` API method for removing loaded estimator from cache. Good for testing changes to the estimator that happen outside the train/deploy scenario, e.g. hyperparameter updates.


montanalow (changes requested):
I think xgboost (actually all algorithm) concurrency should be a global setting for training and inference, rather than a model specific setting.

levkk:
Implemented GUC parameter in #463

PR #445, opened 2022-10-21
Fully qualify table name

levkk:
1. Create tables in the public schema in the dashboard by default
2. Check that the table exists in the public schema or otherwise request fully qualified table name

### DIFF
--- a/pgml-extension/src/orm/snapshot.rs
+++ b/pgml-extension/src/orm/snapshot.rs
@@ -193,6 +193,10 @@ impl Snapshot {
     ) -> Snapshot {
         let mut snapshot: Option<Snapshot> = None;
         let status = Status::in_progress;
+
+        // Validate table exists.
+        let (_schema_name, _table_name) = Self::fully_qualified_table(relation_name);
+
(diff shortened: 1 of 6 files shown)


pgml-extension/src/orm/snapshot.rs
```diff
@@ -256,22 +260,61 @@ impl Snapshot {
...
+            ),
+        };
+
+        match schema_name {
+            None => {
+                let table_count = Spi::get_one_with_args::<i64>("SELECT COUNT(*) FROM information_schema.tables WHERE table_name = $1 AND table_schema = 'public'", vec![
+                    (PgBuiltInOids::TEXTOID.oid(), table_name.clone().into_datum())
+                ]).unwrap();
+
+                let error = format!("Relation \"{}\" could not be found in the public schema. Please specify the table schema, e.g. pgml.{}", table_name, table_name);
```
montanalow, line 286: Why doesn't the search path take care of this?

levkk: `fn analysis`

PR #458, opened 2022-10-25
Dont materialize snapshots by default

levkk:


montanalow (approved):
Cool

montanalow (commented):
Have you tested that the dashboard handles missing snapshots properly? It should since they were always delete-able and not migrated in v1 to v2.

levkk:
> Have you tested that the dashboard handles missing snapshots properly? It should since they were always delete-able and not migrated in v1 to v2.

Good catch, I didn't

PR #481, opened 2022-11-07
Move to pgx 0.5

levkk:
- Moving to `tcdi/pgx` 0.5.6
- Dropping support for Postgres 10 as a result (EOL)
- Maybe fix tests because we're rebuilding the world now


pgml-extension/src/orm/model.rs
```diff
@@ -105,7 +105,7 @@ impl Model {
                   (PgBuiltInOids::TEXTOID.oid(), search.map(|search| search.to_string()).into_datum()),
                   (PgBuiltInOids::JSONBOID.oid(), search_params.into_datum()),
                   (PgBuiltInOids::JSONBOID.oid(), search_args.into_datum()),
-                  (PgBuiltInOids::INT4OID.oid(), dataset.num_features.into_datum()),
```
montanalow, line 108: int4 is not i64 it's i32

levkk: Ah boo

PR #514, opened 2023-01-19
Dashboard meet Rust

levkk:
Rewrote `pgml-dashboard` in Rust (from Python/Django). Fun :)


dusanmarjanovic:
i want to upgrade from django to rust dashboard, but can you please adjust the documentation or just let us know here how should we build and run it? thanks

montanalow:
> i want to upgrade from django to rust dashboard, but can you please adjust the documentation or just let us know here how should we build and run it? thanks

@dusanmarjanovic I've updated the docs in #522. There is now a README.md in the pgml-dashboard directory with build/run/test directions, as well as an update to the [installation docs](https://postgresml.org/user_guides/setup/v2/installation/)

PR #524, opened 2023-02-01
Namespace the migrations under the pgml schema

montanalow:
This is a breaking change, from the last release, and needs review.


levkk:
Pretty sure this breaks migrations. Sqlx doesn't let you edit existing migrations and only allows writing new ones.

levkk (approved):
LGTM

montanalow:
Given we're relatively recent since the original issue was merged, and that there haven't even been build instructions for the Rust dashboard since it was merged that would show how to run these migrations, I'm inclined to think the fallout from a breaking "fix" to avoid colliding with the public schema is worth dealing with the fallout for anyone running master.

There's no way for the dashboard application to enforce that the `_sqlx_migrations` table is moved into the pgml schema currently due to limitations in our upstream sqlx dependency, although there are [open feature](https://github.com/launchbadge/sqlx/issues/1698) requests along these lines. That table will currently end up in the default search_path, wherever that happens to be. For developers, that will be the `public` schema, but deployements can set the default search path for the dashboard user to force it into `pgml` schema as well.

I've consolidated all the docs for application setup into the developer overview, and brought them up to date and inline with the current appropriate values. Most people will need to start there development environment over with this change. :(

Issue #530, opened 2023-02-03
signal 11: Segmentation fault


dusanmarjanovic:
I've pulled fresh version of postgresml from github instead of using ubuntu package (as from what I saw it still doesn't have preprocessing features) and managed to compile it and update .so library with the new one. But now it gives me a segfault even on the basic example:

  2023-02-03 16:34:32.815 UTC [125591] LOG:  database system is ready to accept connections
  2023-02-03 16:34:44.698 UTC [125591] LOG:  server process (PID 125611) was terminated by signal 11: Segmentation fault
  2023-02-03 16:34:44.698 UTC [125591] DETAIL:  Failed process was running: SELECT * FROM pgml.train(
            project_name => 'Breast Cancer Detection',
            task => 'classification',
            relation_name => 'pgml.breast_cancer',
            y_column_name => 'malignant'
          )

ubuntu 22.04, postgresql 14

montanalow:
I'm guessing you can successfully run tests locally on your build with `cargo pgx test`? Did you build w/ `cargo pgx package`? This sounds like there may be some incompatibility between different parts of the installation, maybe something is leftover from the previous `create extension ...` call? Have you tried doing a `cargo pgx install` (which depends on `pg_config` being setup with the correct paths)?

Can you add the exact steps you went through to upgrade the .so?

PS. We're planning to cut a new release with preprocessing next week.

dusanmarjanovic:
all tests run successfully:
test result: ok. 47 passed; 0 failed; 0 ignored; 0 measured; 0 filtered
out; finished in 279.30s

Yes, built with cargo pgx package.
For the steps, I just followed the exact instructions on building from
source and just copied the compiled .so instead of the one that came from
the .deb package.

Tried now with cargo pgx install also, it was done successfully but still
the same seg fault. If there's anything else I can provide please let me
know, otherwise I'll have to wait for the prepared package.

montanalow:
This is interesting. I can't repro any segfaults on my setup, but I can get one on our build machine during `cargo pgx package`. It's coming from the installation process during pgx macros generating function signatures. Why this works on some machines and not others is not obvious to me. I'm going to work on a patch for the build process, including updating pgx, before doing an actually release (that includes preprocessing).

dusanmarjanovic:
Thanks, if you need me to test that patch before releasing let me know, also this is my test vm running nothing but pgml so I can give you access if you need it for some testing etc.

montanalow:
[...]

PR #533, opened 2023-02-07
update to pgx 0.7.11

montanalow:
Investigate #530


montanalow:
@levkk Tests pass on my local and the build server. Can you see if you can still repro a segfault during build on your local?

levkk (approved):
Lgtm, will check out the segfault tomorrow

PR #552, opened 2023-02-21
Dashboard snapshots fix

santiatpml:
Snapshots tab will only show tables/views that are present in the live table.

### DIFF
--- a/pgml-dashboard/src/models.rs
+++ b/pgml-dashboard/src/models.rs
@@ -549,8 +549,7 @@ impl Snapshot {
                     analysis,
                     created_at,
                     updated_at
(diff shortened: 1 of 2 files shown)


pgml-dashboard/src/models.rs
```diff
@@ -549,8 +549,7 @@ impl Snapshot {
                     analysis,
                     created_at,
                     updated_at
-                FROM pgml.snapshots
-            "
+                FROM pgml.snapshots JOIN pg_class ON oid::regclass::text = relation_name"
```
montanalow, line 552: Does this work with views as well as tables?

santiatpml: Yes, it was tested on both views and tables.

montanalow (approved):

PR #553, opened 2023-02-23
Santi dashboard snapshots fix2

santiatpml:
Reference from models to snapshots is fixed.


pgml-dashboard/src/guards.rs
```diff
@@ -49,7 +49,10 @@ impl<'r> FromRequest<'r> for Cluster {
             None => return Outcome::Failure((Status::BadRequest, ())),
         };
 
-        Outcome::Success(Cluster { pool, context: shared_state.get_context(cluster_id) })
+        Outcome::Success(Cluster {
+            pool,
+            context: shared_state.get_context(cluster_id),
+        })
```
montanalow, line 55: Looks like we need to add a `cargo fmt` to the ci suite. cc @levkk

PR #554, opened 2023-02-23
Santi dashboard snapshots fix2

santiatpml:
Removed debug messages.


pgml-dashboard/src/lib.rs
```diff
@@ -397,9 +400,13 @@ pub async fn models_index(cluster: Cluster) -> Result<ResponseOk, errors::Error>
 #[get("/models/<id>")]
 pub async fn models_get(cluster: Cluster, id: i64) -> Result<ResponseOk, errors::Error> {
     let model = models::Model::get_by_id(cluster.pool(), id).await?;
-    let snapshot = models::Snapshot::get_by_id(cluster.pool(), model.snapshot_id).await?;
+    let mut snapshot = models::Snapshot::get_by_id(cluster.pool(), model.snapshot_id).await?;
     let project = models::Project::get_by_id(cluster.pool(), model.project_id).await?;
 
+    if !snapshot.relation_exists(cluster.pool()).await? {
+        snapshot.relation_name = String::from("");
```
montanalow, line 407: I think an additional boolean attribute representing "deleted" rather than blanking the original relation name might be useful, and then we can additionally present this state in the UX. We can also cache the `relation_exists` logic for use in multiple places this way. 

Expanding on the concept that a materialized snapshot may be deleted if it's actual data is deleted, and a non materialized snapshot is deleted if the table is dropped.

montanalow (changes requested):

pgml-dashboard/src/lib.rs
santiatpml: Snapshots has a new boolean field `exists` along with `table_size`. They are part of SQL queries now instead of standalone functions.

pgml-dashboard/test_table.sql~
```diff
@@ -0,0 +1,9 @@
+
+SELECT id,relation_name,y_column_name,
```
montanalow, line 2: I think this file may have been added by mistake?

montanalow (approved):
This looks like a solid improvement to me!

pgml-dashboard/test_table.sql~
santiatpml: Yep, will get rid of it and make the frontend changes that Lev recommended and push it.

PR #556, opened 2023-02-24
Dashboard backend tests using rocket

santiatpml:
Tests for all index pages.


pgml-dashboard/src/main.rs
```diff
@@ -29,3 +29,6 @@ async fn main() {
         .await
         .expect("failed to shut down Rocket");
 }
+
+
+mod tests;
```
levkk, line 34: Make sure to run `cargo fmt` :)

santiatpml: Done.

pgml-dashboard/src/main.rs
```diff
@@ -29,3 +29,5 @@ async fn main() {
         .await
         .expect("failed to shut down Rocket");
 }
+
+mod tests;
```
levkk, line 33: You can put the tests in the same file as the code being tested. This is the Rust way, tests and code together. Rust also has integration tests, which go into a separate app that then uses public methods of the crate it's testing.

pgml-dashboard/src/tests.rs
```diff
@@ -0,0 +1,67 @@
+use pgml_dashboard::Clusters;
+use pgml_dashboard::{index, migrate, paths};
+use rocket::fs::FileServer;
+use rocket::{Build, Rocket};
+
+async fn rocket() -> Rocket<Build> {
```
levkk, line 6: You can move this function inside the `test` module because it's the only place where it's being used.

pgml-dashboard/src/main.rs
santiatpml: Moved tests to main.rs

pgml-dashboard/src/tests.rs
santiatpml: Moved rocket() into tests.

levkk (approved):
:100:

PR #557, opened 2023-02-27
Dashboard backend tests

santiatpml:
Using scraper to get all relevant `hrefs` on each page and checking for status code 200.


pgml-dashboard/Cargo.toml
```diff
@@ -24,3 +24,4 @@ bigdecimal = "0.3"
 chrono = "0.4"
 serde_json = "1"
 csv-async = "1"
+scraper = "0.14.0"
```
levkk, line 27: It would be good to enable automatic new lines at the end of the files. For VsCode, I found [this](https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files).

pgml-dashboard/src/main.rs
```diff
@@ -95,5 +109,79 @@ mod test {
...
-}
 
+    #[rocket::async_test]
+    async fn test_snapshot_entries() {
+        let snapshots_endpoint = "/dashboard/snapshots/";
+        let client = Client::tracked(rocket().await).await.unwrap();
+        let response = client.get(snapshots_endpoint).dispatch().await;
+
+        let body = response.into_string().await.unwrap();
+        let snapshot_links = get_href_links(body.as_str(), snapshots_endpoint);
```
levkk, line 120: This will follow the links in the menu I think, so it'll double test those pages, which is totally fine since it's fast.

levkk (approved):
LGTM!

pgml-dashboard/src/main.rs
santiatpml: All those links are dropped in `get_href_links` function here 

```
if href.contains(pattern) && href != pattern {
                output.push(String::from(href));
}
```

PR #561, opened 2023-03-11
fix vscode error

chillenberger:


pgml-dashboard/Cargo.toml
```diff
@@ -14,7 +14,7 @@ include = ["src/", "sqlx-data.json", "templates/", "migrations/", "static/"]
 [dependencies]
 sailfish = "0.5.0"
 rocket = {version = "0.5.0-rc.2", features = ["secrets"] }
-sqlx = { version = "0.6", features = [ "runtime-tokio-rustls", "postgres", "json", "tls", "migrate", "time", "uuid", "bigdecimal", "offline",] }
```
levkk, line 17: Could you also delete `pgml-dashboard/sqlx-data.json`? That's what `offline` is used for, and since we're removing this feature, we don't need that file anymore.

levkk (approved):
LGTM!

PR #566, opened 2023-03-15
disambiguate api

### LINKED ISSUE #549: version 2.2.0 gives pgml.predict is not unique
New version installed from the apt repo gives error such as:
error returned from database: function pgml.predict(unknown, smallint[]) is not unique

Caused by:
    function pgml.predict(unknown, smallint[]) is not unique

even for the most of the examples in the notebooks section.

montanalow:
Resolves #549


dusanmarjanovic:
i just built the new one and now when i try the example i get segfault on two different environments?
2023-03-25 17:37:54.692 UTC [47842] LOG:  server process (PID 47864) was terminated by signal 11: Segmentation fault
2023-03-25 17:37:54.692 UTC [47842] DETAIL:  Failed process was running: SELECT * FROM pgml.train('Handwritten Digits', 'classification', 'pgml.digits', 'target')

PR #578, opened 2023-03-28
Create embedding API

montanalow:
Expose sentence_transformers to generate embeddings from `TEXT`.

### DIFF
--- a/pgml-extension/Dockerfile
+++ b/pgml-extension/Dockerfile
@@ -32,7 +32,7 @@ RUN apt update && apt-fast install -y \
 	python3-pip \
 	libpython3.10-dev \
 	python3.10-dev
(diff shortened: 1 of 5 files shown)


santiatpml (commented):
Is `project` in [Line 55](https://github.com/postgresml/postgresml/blob/611d31e4a9ea22869fa24705b6b5e4185ce9abb2/pgml-extension/src/bindings/transformers.py#L55) same as model from this list? https://www.sbert.net/docs/pretrained_models.html

montanalow:
It's actually this list: https://huggingface.co/models?library=sentence-transformers

I need to rename the param to something other than project, since it's not. Originally I had a `load_model()` that created the project, rather than using it directly, but that is another can of worms.

PR #579, opened 2023-03-28
More Descriptive code snippet, more clearer comments

yash-gupta167:
In this updated code snippet, the parameter ```test_sampling``` has been replaced with the more descriptive ```test_data_selection```. Additionally, the comments have been updated to provide a clearer explanation of each parameter. This should help make the code more readable and reduce the chances of errors.


montanalow:
I think it would be better to update the documentation, rather than updating the parameter name. Updating the parameter name would break backward compatibility for existing deployments.

PR #588, opened 2023-04-05
update dependencies

montanalow:
New updates for pgx.


montanalow:
Everyone will need to run `cargo install cargo-pgx --version="0.7.4"` to run `cargo pgx` locally.

PR #589, opened 2023-04-07
fix for np.float32 serialization

santiatpml:
Added tests for transforms and moved tune to a different set.

### DIFF
--- a/pgml-extension/Dockerfile.local
+++ b/pgml-extension/Dockerfile.local
@@ -11,7 +11,7 @@ RUN cat /etc/apt/sources.list
 RUN apt-get update && apt-get install -y postgresql-pgml-14
 
 # Cache this, quicker
-RUN pip3 install xgboost scikit-learn diptest torch lightgbm transformers datasets sentencepiece sentence_transformers sacremoses sacrebleu rouge protobuf
+RUN pip3 install xgboost scikit-learn diptest torch lightgbm transformers datasets sentencepiece sentence_transformers sacremoses sacrebleu rouge
 
 COPY --chown=postgres:postgres . /app
 WORKDIR /app
--- a/pgml-extension/Dockerfile.local
+++ b/pgml-extension/Dockerfile.local
@@ -11,7 +11,7 @@ RUN cat /etc/apt/sources.list
 RUN apt-get update && apt-get install -y postgresql-pgml-14
 
 # Cache this, quicker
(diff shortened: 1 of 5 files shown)


pgml-extension/tests/test.sql
```diff
@@ -27,3 +27,4 @@ SELECT pgml.load_dataset('wine');
 \i examples/multi_classification.sql
 \i examples/regression.sql
 \i examples/vectors.sql
+\i examples/transformers.sql
```
montanalow, line 30: We don't run transformers or finetuning during tests, since they take a lot of RAM and download time.

montanalow (changes requested):

pgml-extension/tests/test.sql
santiatpml: Removed them.

montanalow (approved):

pgml-extension/Dockerfile.local
```diff
@@ -11,7 +11,7 @@ RUN cat /etc/apt/sources.list
 RUN apt-get update && apt-get install -y postgresql-pgml-14
 
 # Cache this, quicker
-RUN pip3 install xgboost scikit-learn diptest torch lightgbm transformers datasets sentencepiece sentence_transformers sacremoses sacrebleu rouge
+RUN pip3 install xgboost scikit-learn diptest torch lightgbm transformers datasets sentencepiece sentence_transformers sacremoses sacrebleu rouge protobuf
```
montanalow, line 14: we should not need protobuf, what requires that?

santiatpml: Finbert model needs protobuf.
```sql
SELECT pgml.transform(
    inputs => ARRAY[
        'Stocks rallied and the British pound gained.', 
        'Stocks making the biggest moves midday: Nvidia, Palantir and more'
    ],
    task => '{"task": "text-classification", 
              "model": "ProsusAI/finbert"
             }'::JSONB
) AS market_sentiment;
```

montanalow: wow... that's... indicative of a whole new level... are they actually calling any rpcs, or are they just serializing in a round about way? I don't really care...

santiatpml: I could not reproduce the error I saw before without protobuf on two other machines. Looks like we don't need it.

PR #593, opened 2023-04-16
Adds pipeline model caching in the `transform` function.

f-prime:
This PR adds the ability to run a query that looks like this:

```
SELECT pgml.transform(
        '{"model": "roberta-large-mnli"}'::JSONB, 
        inputs => ARRAY[
            'I love how amazingly simple ML has become!', 
            'I hate doing mundane and thankless tasks. ☹️'
        ],
        cache => TRUE
    ) AS positivity;
```

`roberta-large-mnli` will be cached in memory to prevent `transformers.pipeline()` from being called more than once for the same model.

By default, `cache` is `FALSE`.

### DIFF
--- a/pgml-extension/src/bindings/transformers.py
+++ b/pgml-extension/src/bindings/transformers.py
@@ -39,6 +39,7 @@ from transformers import (
 
 __cache_transformer_by_model_id = {}
 __cache_sentence_transformer_by_name = {}
+__cache_transform_pipeline_model_by_name = {}
 
 class NumpyJSONEncoder(json.JSONEncoder):
     def default(self, obj):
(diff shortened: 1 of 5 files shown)


pgml-extension/src/bindings/transformers.py
```diff
@@ -39,19 +39,26 @@
...
         return super().default(obj)
 
-def transform(task, args, inputs):
+def transform(task, args, inputs, cache_model):
     task = json.loads(task)
     args = json.loads(args)
     inputs = json.loads(inputs)
 
-    pipe = transformers.pipeline(**task)
+    model = task.get("model")
```
montanalow, line 55: I think there may be different pipelines with the same model, to handle different tasks. We may need to use the full task param for caching. Something like:

```
    if cache:
        key = ','.join([str(key) + ':' + str(value) for (key, value) in sorted(task.items())])
        if key not in __cache_transformer_by_task:
            __cache_transformer_by_task[key] = transformers.pipeline(**task)
        pipe = __cache_transformer_by_task[key]
    else:
        pipe = transformers.pipeline(**task)
```

f-prime: I see, good point. I'll think some more about the parameters of .pipeline() and push an update.

f-prime: Okay, I pretty much copied your code verbatim. I also like the idea in there where you have to pass the `cache` parameter also to USE the cached model. My initial version always used the cached version if it was available whether or not the cache flag was passed in or not.

montanalow:
This looks good to me as an improvement over the status quo, so I'm going to merge it. We're being slightly inconsistent, i.e.  every other API caches by default, including the `pgml.embed()` which may use similarly large models. I'm not really content with that approach either since the only way to clear those caches is to drop the connection and kill the backend process. It's good to have these examples though, that we'll want to design a consistent interface around for more feature rich cache management.

PR #615, opened 2023-05-04
Cache miss

### LINKED ISSUE #616: Friendly error for HuggingFace models that don't exist
```
select pgml.embed('intfloat/e5_large', 'this is a test');

INFO:  Cache miss for "intfloat/e5_large", loading transformer, please wait
ERROR:  called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'huggingface_hub.utils._errors.RepositoryNotFoundError'>, value: RepositoryNotFoundError('401 Client Error. (Request ID: Root=1-6453dc33-0113b3d85bfa81fa2711c0f2)\n\nRepository Not Found for url: https://huggingface.co/api/models/intfloat/e5_large.\nPlease make sure you specified the correct `repo_id` and `repo_type`.\nIf you are trying to access a private or gated repo, make sure you are authenticated.\nInvalid username or password.'), traceback: Some(<traceback object at 0x7f72f59c4ec0>) }
```

What actually happened here is the transformer name is with a dash, not an underscore. We should return a friendlier error that tells the user what actually happened.

levkk:
Fix #616

### DIFF
--- a/pgml-extension/src/bindings/transformers.rs
+++ b/pgml-extension/src/bindings/transformers.rs
@@ -62,13 +62,17 @@ pub fn embed(transformer: &str, text: &str, kwargs: &serde_json::Value) -> Vec<f
     let kwargs = serde_json::to_string(kwargs).unwrap();
     Python::with_gil(|py| -> Vec<f32> {
         let is_embed_cached: Py<PyAny> = PY_MODULE.getattr(py, "is_embed_cached").unwrap().into();
-        let cached: bool  = is_embed_cached.call1(
-            py,
-            PyTuple::new(py, &[transformer.to_string().into_py(py)])
-        ).unwrap().extract(py).unwrap();
+        let cached: bool = is_embed_cached
+            .call1(py, PyTuple::new(py, &[transformer.to_string().into_py(py)]))
+            .unwrap()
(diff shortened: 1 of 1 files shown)


pgml-extension/src/bindings/transformers.py
```diff
@@ -92,6 +92,11 @@ def embed(transformer, text, kwargs):
     return result
 
 
+def is_embed_cached(transformer):
```
montanalow, line 95: I think it'd be better to move this inside the python embed function and raise an exception if it's not cached, so we aren't grabbing the GIL from Rust twice. In the exception case, we can log, call load, and retry.

montanalow (changes requested):

pgml-extension/src/bindings/transformers.py
levkk: Raise an exception? We don't want to crash `pgml.embed`...

santiatpml: Do `print` statements inside transformers.py display while running `pgml.embed`?

montanalow: No, you need to go back to rust to print.

Issue #617, opened 2023-05-04
Functions using `blas` cause a segfault (SIGSEV)


eeeebbbbrrrr:
After working with @levkk and @montanalow to install PostgresML (as of master: 63ebce31ef599d17c966c750847f7d23dd0754b9) on my linux box, I discovered that functions such as `pgml.cosine_similarity` and `pgml.norm_l1` cause Postgres to segfault.

As an example:

```sql
[v15.1][5126] pgml=# select pgml.norm_l1(ARRAY[1,2,3]::real[]);
server closed the connection unexpectedly
    This probably means the server terminated abnormally
    before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
The connection to the server was lost. Attempting reset: Failed.
Time: 188.973 ms
[v][] ?!> 
```

Postgres logs leading up to a crash against `pgml.cosine_similarity()` are:

```
/usr/local/lib/python3.10/dist-packages/torch/distributed/elastic/utils/logging.py:65: RuntimeWarning: Error deriving logger module name, using <None>. Exception: <module '' from '/home/pg/15/data'> is a built-in module
  warnings.warn(
No sentence-transformers model found with name /home/zombodb/.cache/torch/sentence_transformers/intfloat_e5-large. Creating a new one with MEAN pooling.
2023-05-04 18:35:48.950 UTC [20973] LOG:  server process (PID 21218) was terminated by signal 11: Segmentation fault
2023-05-04 18:35:48.950 UTC [20973] DETAIL:  Failed process was running: select *, pgml.cosine_similarity(embed, pgml.embed('intfloat/e5-large', 'meetings with beer or wine and cheese')) from embeddings_e5large_100k limit 10;
2023-05-04 18:35:48.950 UTC [20973] LOG:  terminating any other active server processes
2023-05-04 18:35:48.953 UTC [20973] LOG:  all server processes terminated; reinitializing
2023-05-04 18:35:48.979 UTC [20973] FATAL:  Can't attach, lock is not in an empty state: PgLwLockInner
2023-05-04 18:35:48.980 UTC [20973] LOG:  database system is shut down
```

The backtrace from a --debug build of pgml is:

```
Thread 1 "postgres" received signal SIGSEGV, Segmentation fault.
0x00007ff52bc65d76 in sdot_ () from /home/pg/15/lib/postgresql/pgml.so
(gdb) bt
[...]

PR #620, opened 2023-05-04
remove static linking

montanalow:
#617 fix

### DIFF
--- a/pgml-extension/tests/test.sql
+++ b/pgml-extension/tests/test.sql
@@ -27,4 +27,5 @@ SELECT pgml.load_dataset('wine');
 \i examples/multi_classification.sql
 \i examples/regression.sql
 \i examples/vectors.sql
-
+-- transformers are generally too slow to run in the test suite
+--\i examples/transformers.sql
--- a/pgml-extension/build.rs
+++ b/pgml-extension/build.rs
@@ -1,3 +1,0 @@
-fn main() {
-    println!("cargo:rustc-link-lib=static=openblas");
-}
--- a/pgml-extension/examples/image_classification.sql
+++ b/pgml-extension/examples/image_classification.sql
@@ -65,6 +65,12 @@ SELECT * FROM pgml.train('Handwritten Digits', algorithm => 'xgboost', hyperpara
 -- Histogram Gradient Boosting is too expensive for normal tests on even a toy dataset
 -- SELECT * FROM pgml.train('Handwritten Digits', algorithm => 'hist_gradient_boosting', hyperparams => '{"max_iter": 2}');
 
+-- runtimes
+SELECT * FROM pgml.train('Handwritten Digits', algorithm => 'linear', runtime => 'python');
+SELECT * FROM pgml.train('Handwritten Digits', algorithm => 'linear', runtime => 'rust');
+
+SELECT * FROM pgml.train('Handwritten Digits', algorithm => 'xgboost', runtime => 'python', hyperparams => '{"n_estimators": 10}');
+SELECT * FROM pgml.train('Handwritten Digits', algorithm => 'xgboost', runtime => 'rust', hyperparams => '{"n_estimators": 10}');
 
 -- check out all that hard work
 SELECT trained_models.* FROM pgml.trained_models 
--- a/pgml-extension/examples/regression.sql
+++ b/pgml-extension/examples/regression.sql
@@ -85,6 +85,12 @@ SELECT * FROM pgml.train('Diabetes Progression', algorithm => 'xgboost', hyperpa
 -- Histogram Gradient Boosting is too expensive for normal tests on even a toy dataset
 -- SELECT * FROM pgml.train('Diabetes Progression', algorithm => 'hist_gradient_boosting', hyperparams => '{"max_iter": 10}');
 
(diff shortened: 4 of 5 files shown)


pgml-extension/tests/test.sql
```diff
@@ -27,4 +27,5 @@ SELECT pgml.load_dataset('wine');
 \i examples/multi_classification.sql
 \i examples/regression.sql
 \i examples/vectors.sql
-
+-- transformers are generally too slow to run in the test suite
+--\i examples/transformers.sql
```
montanalow, line 31: @levkk can you run tests in this file including transformers on AMD? You'll need to uncomment this line.

```
psql -h localhost -p 28815 -d pgml -f tests/test.sql -P pager
```

levkk:
These issues are intermittent so until we understand why they happen, we won't have any "fix", just attempts at fixing it.

pgml-extension/tests/test.sql
levkk: Even if it passes, it won't necessarily prove we fixed anything, since we added that line explicitly to "fix" linking issues we've had in the past. Maybe we should just move openblas to be dynamically linked? We will need to fork the crate and change the option.

montanalow: Is there any reference to what "linking issues" we've had? A lot has changed since this was introduced (maybe even including our linker).

levkk: Yeah, we would segfault inside scikit when trying to train a logistic regression model (or maybe it was a linear regression). One of those uses clbas.

montanalow:
We default to Rust now for linear regression.

pgml-extension/tests/test.sql
montanalow: I added some more tests. No issues on intel specifying the runtime as python or rust for both linear and logistic regression, or xgboost.

montanalow: I think what may have been happening a year ago, was that scikit actually has had many bugs with CBLAS, which is our preferred link target. My guess is crashes were happening on machines with old versions of scikit installed. They largely removed CBLAS in favor of cython blas in [2019](https://github.com/scikit-learn/scikit-learn/pull/13203).

levkk: We had a newer version of Scikit than 2019.

montanalow: Ubuntu 20.04LTS Focal was using an [older unpatched version of scikit](https://packages.ubuntu.com/source/focal/scikit-learn), which likely would have been popular when we started developing this package. I think we should go back to the default dynamic linking behavior, although I'd love to dig further into @eeeebbbbrrr's setup to understand which blas dependency is installed on his machine:

On my machine: ```
$ apt show libblas-dev -a
Package: libblas-dev
Version: 3.10.0-2ubuntu1
Priority: optional
Section: libdevel
Source: lapack
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Debian Science Team <debian-science-maintainers@lists.alioth.debian.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 1084 kB
Provides: libblas.so
Depends: libblas3 (= 3.10.0-2ubuntu1)
Suggests: liblapack-doc
Homepage: https://www.netlib.org/lapack/
Download-Size: 164 kB
APT-Sources: http://archive.ubuntu.com/ubuntu jammy/main amd64 Packages
Description: Basic Linear Algebra Subroutines 3, static library
 This package is a binary incompatible upgrade to the blas-dev
 package. Several minor changes to the C interface have been
 incorporated.
 .
 BLAS (Basic Linear Algebra Subroutines) is a set of efficient
 routines for most of the basic vector and matrix operations.
 They are widely used as the basis for other high quality linear
 algebra software, for example lapack and linpack.  This
 implementation is the Fortran 77 reference implementation found
 at netlib.
 .
 This package contains a static version of the library.
 ```

montanalow:
Funny enough, google only knows 1 source for this config, which happens to be @levkk's blog post, which documents that I was the original reporter of a crash, that I'd completely forgotten about, and hadn't really understood, because he was the one doing the Python/Rust transition at the time.

https://postgresml.org/blog/backwards-compatible-or-bust-python-inside-rust-inside-postgres 

We know that old versions of scikit crash using CBLAS, and I know that I was on a 18.04LTS when this post was written, so it makes sense I started exercising this crash when @levkk gave Python in Rust access to my dynamically linked CBLAS installation with py03, but my system Python was not otherwise configured with access to CBLAS, so it was unaffected.

This line forces openblas-src crate to skip the ["cblas", "system"] features which dynamic links to system CBLAS, and instead statically link to the bundled _OpenBLAS_. This prevents scikit from having access to a dynamically linked CBLAS crash in older versions, which "fixed" my original issue, but now we're seeing issues w/ statically linked OpenBLAS on @eeeebbbbrrrr's machine. I think the "right fix" is to upgrade scikit to a version new than 2019, and dynamically link CBLAS, which is the standard, not to statically link OpenBLAS.

levkk (approved):
I smell another blog post incoming.

PR #622, opened 2023-05-06
4x speedup

montanalow:
Using Array rather than Vec prevents an allocation, and speeds up results of `cosine_similarity` 4x for large scans, putting it within a few percentage points of the vectorized C equivalent code.


montanalow:
@eeeebbbbrrrr This one gets the c equivalent vectorization speedups with 0 copy arrays for most cases, except for aggregates. Those need a bit more work.

PR #636, opened 2023-05-19
pgml Python SDK with vector search support

santiatpml:
Objective of this SDK is to provide an easy interface for PostgresML generative AI capabilities. This version supports vector search using multiple models and text splitters.

Quick start instructions are [here](https://github.com/postgresml/postgresml/blob/santi-pgml-memory-sdk-python/pgml-sdks/python/pgml/README.md)


solidsnack (approved):

pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -0,0 +1,771 @@
...
+        This function creates a table and indexes for storing documents in a PostgreSQL database.
+        """
+        self.documents_table = self.name + ".documents"
+        conn = self.pool.getconn()
+
+        create_schema_statement = "CREATE SCHEMA IF NOT EXISTS %s" % self.name
+        run_create_or_insert_statement(conn, create_schema_statement)
+        create_table_statement = (
+            "CREATE TABLE IF NOT EXISTS %s (\
+                                id          serial8 PRIMARY KEY,\
```
levkk, line 68:
```suggestion
                                id          BIGSERIAL PRIMARY KEY,\
```

More idiomatic, but doesn't matter.

pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -0,0 +1,771 @@
...
+                                created_at  timestamptz NOT NULL DEFAULT now(),\
+                                document    uuid NOT NULL,\
+                                metadata    jsonb NOT NULL DEFAULT '{}',\
+                                text        text NOT NULL,\
+                                UNIQUE (document)\
```
levkk, line 73: Duplicate primary key technically, since this is unique. Curious why we can't use the `id` field as the document identifier?

pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -0,0 +1,771 @@
...
+        create_index_statement = (
+            "CREATE INDEX CONCURRENTLY IF NOT EXISTS \
+                                    created_at_index ON %s \
+                            (created_at);"
+            % (self.documents_table)
+        )
+        run_create_or_insert_statement(conn, create_index_statement, autocommit=True)
+
+        create_index_statement = (
+            "CREATE INDEX CONCURRENTLY IF NOT EXISTS \
```
levkk, line 88: Don't need to create an index here on `document`, `UNIQUE` does it already automatically. So I think you end up with two indexes on the same field.

pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -0,0 +1,771 @@
...
+                              ON DELETE CASCADE\
+                              ON UPDATE CASCADE\
+                              DEFERRABLE INITIALLY DEFERRED,\
+                            UNIQUE (task, splitter, model)\
+                    );"
+            % (self.transforms_table, self.splitters_table, self.models_table)
+        )
+        run_create_or_insert_statement(conn, create_statement)
+
+        index_statement = (
```
levkk, line 208: `UNIQUE (task, splitter, model)` in the table definition creates a compound index on those three columns. Having 3 more individual indexes on the same columns may not be necessary.

pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -0,0 +1,771 @@
...
+        """
+        conn = self.pool.getconn()
+        self.transforms_table = self.name + ".transforms"
+        create_statement = (
+            "CREATE TABLE IF NOT EXISTS %s (\
+                            oid         regclass PRIMARY KEY,\
+                            created_at  timestamptz NOT NULL DEFAULT now(), \
+                            task        text NOT NULL, \
+                            splitter    int8 NOT NULL REFERENCES %s\
+                              ON DELETE CASCADE\
```
levkk, line 195: Are you sure you want to `CASCADE`? This has the effect of automatically deleting rows from this table when the row they are referencing in another table is deleted. This can delete a lot of data accidentally. The preferred way for me generally is to `ON DELETE RESTRICT` which is the default. That way, you'll get an error when attempting to delete a splitter that's referenced by this table. This is default behavior also, so you can just remove `ON DELETE CASCADE`.

pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -0,0 +1,771 @@
...
+
+    def _create_transforms_table(self) -> None:
+        """
+        This function creates a transforms table in a PostgreSQL database and adds indexes to it.
+        """
+        conn = self.pool.getconn()
+        self.transforms_table = self.name + ".transforms"
+        create_statement = (
+            "CREATE TABLE IF NOT EXISTS %s (\
+                            oid         regclass PRIMARY KEY,\
```
levkk, line 191: Odd way of referencing another table, I've never seen that done before. Is this compatible with logical replication? I.e. if we want to move this data to another database, the oids probably won't match anymore, since they are specific to a Postgres installation.

pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -0,0 +1,771 @@
...
+            embeddings_size = len(
+                self._get_embeddings(
+                    conn=conn, model_name=model_name, parameters=model_parameters
+                )
+            )
+            create_statement = (
+                "CREATE TABLE IF NOT EXISTS %s ( \
+                                id          serial8 PRIMARY KEY,\
+                                created_at  timestamptz NOT NULL DEFAULT now(),\
+                                chunk       int8 NOT NULL REFERENCES %s\
```
levkk, line 602: The convention is sometimes to use `field_id` to explain to the reader that this refers to the primary key (usually `id`) column in another table. The reviewer would look in this case if the column being referenced has an index on it, which is often required for foreign keys: a foreign key validation needs to be an index scan, the quickest way to validate that this value exists in another table.

levkk (approved):
The tables feel a bit overindexed. In Postgres, each index carries a write penalty: for each row that's inserted into the table, each index needs to be updated accordingly. When all (or most) columns are updated, an update to any column requires an update to all indexes (i.e. Postgres can't do HOT, heap-only, updates).

I think I understand the schema overall, although it would be helpful to define it in a separate `.sql` file which can then be executed when the SDK is used for the first time. Although, we do generate a lot of tables in the fly, so understandably that's not possible for all use cases.

Overall, I think this is great to start with, and might require some optimizations as it's deployed at scale.

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: I think it's better to stick with the `*4`, `*8` variants -- they are more consistent. See https://www.postgresql.org/docs/7.4/datatype.html#DATATYPE-TABLE

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: The customer can supply the UUID, is my understanding -- they both have to be unique because this is where we correlate internal and external IDs.

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: In these tables, input is write-once and additional rows in other tables are generated formulaically from the initial data in a repeatable way. The deletion path is invoked to remove a write-once record and everything associated with it -- it's a kind of GC, and that is basically what `CASCADE` is for.

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: That is Hungarian notation and a Rails convention. The `REFERENCES` clause is the only reliable guide to foreign keys.

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: There is some documentation about this but OID is not the same as the designated registered types -- `regclass`, `regrole` and so on.

solidsnack:
It is much better to improve performance later by removing indexes than to try to improve it by adding them...I'm not sure how this kind of thinking about indexes got started.

levkk:
> It is much better to improve performance later by removing indexes than to try to improve it by adding them...I'm not sure how this kind of thinking about indexes got started.

Experience. Removing an index is dangerous, adding an index is safe.

solidsnack:
This depends on many factors. Removing an unused index requires an exclusive lock; this is not intrinsically a long running operation. Adding a needed index on a large dataset frequently means a lot of additional database activity for an extended period of time.

montanalow (changes requested):

pgml-sdks/python/pgml/pgml/database.py
```diff
@@ -0,0 +1,93 @@
...
+        else:
+            insert_statement = "INSERT INTO pgml.collections (name) VALUES ('%s')" % (
+                name
+            )
+            run_create_or_insert_statement(conn, insert_statement)
+            # Create collection object
+        self.pool.putconn(conn)
+        return Collection(self.pool, name)
+
+    def delete_collection(self, name: str) -> None:
```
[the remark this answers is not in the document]
solidsnack: It is the only way to actually free storage space. Maybe delete should be made two phase?

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: It's like writing `document_ptr` for a pointer.

The only reliable way to know if something is a foreign key or not and thus needs an index is if it says `REFERENCES` -- I'm not sure we want a style of review that attaches to something non-essential in place of that.

pgml-sdks/python/pgml/pgml/database.py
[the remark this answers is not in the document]
montanalow: Yeah, we could have some archiving mechanism, which simply alters the schema to `name` + '_archive', and then enable delete or restore on archives.

pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -0,0 +1,771 @@
...
+                    "%s in not present, skipping this document..." % text_key
+                )
+                continue
+            if id_key in list(document.keys()):
+                document_id = document.pop(id_key)
+            else:
+                log.info("id key is not present.. hashing")
+                document_id = hashlib.md5(text.encode("utf-8")).hexdigest()
+            metadata = document
+            delete_statement = "DELETE FROM %s WHERE document = %s" % (
```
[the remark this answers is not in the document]
solidsnack: The parameter that is being passed is literally called `document_id`.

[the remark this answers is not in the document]
montanalow: Right, so using one convention for names at one layer, and a different convention for names in the next layer down seems confusing, there is good reason to add `_id` in both layers.

pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -0,0 +1,771 @@
...
+
+        results = run_select_statement(conn, select_statement)
+        for result in track(results, description="Generating chunks"):
+            log.debug("Came into chunk insertion")
+            document = result["id"]
+            text = result["text"]
+            chunks = text_splitter.create_documents([text])
+            for chunk_id, chunk in enumerate(chunks):
+                insert_statement = (
+                    "INSERT INTO %s (document,splitter,chunk_id, chunk) VALUES (%s, %s, %s, %s);"
```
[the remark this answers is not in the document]
solidsnack: There seems to be something else going on here because it looks like we store an `id` and a `chunk_id` and the chunk text and this is the chunks table. Is the `chunk_id` actually extraneous?

pgml-sdks/python/pgml/pgml/collection.py
levkk: A foreign key can be polymorphic, in which case, the naming convention suddenly becomes important. Sure, one can't really enforce naming conventions programmatically, but there is no reason to make something harder to understand.

For example, `created_at` can be renamed to `random_date` and only by looking at the definition of the column one would realize that this is actually a creation timestamp, but we don't do that, we call it `created_at` because we want to be kind to our future selves and to reviewers and other users of our code. An argument can be made for calling variables `abcd` and only by reading the code fully, one will truly understand what they do, but we don't do that either.

Rails can be many things that we don't agree with, but the way ActiveRecord handles data is fantastic, and we would be wise to learn from their experience and do something most engineers are familiar and happy to work with.

montanalow: And you will commonly see code that is forced to name two variables in a context things like `document_ptr` and a `document` , to disambiguate the one that has been dereferenced safely. No one would assume that the compiler checks `_ptr` correctness, but if they ever saw a variable with that extension, next to a variable without the extension, and the truth of the references disagreed with the conventions we've established match expectations, people would say the variable was misnamed and likely to cause bugs or confusion.

pgml-sdks/python/pgml/pgml/database.py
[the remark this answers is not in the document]
levkk: Adding more storage space is cheap. Recovering lost data is expensive and scary.

pgml-sdks/python/pgml/pgml/collection.py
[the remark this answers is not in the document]
santiatpml: Yes, `chunk_id` is the chunk index for the given document text. For example: if the text "hello world" is split into two chunks then chunk_id 0 will map to "hello" and chunk_id 1 will map to "world". `id` is the global id of the chunk across all documents and splitters.

pgml-sdks/python/pgml/pgml/collection.py
levkk: This should probably be called `table_name`. An `oid` is something very Postgres-specific and does not mean a table reference at all. In fact,it can reference a row in a TOAST table or a type in `pg_type`.

pgml-sdks/python/pgml/pgml/database.py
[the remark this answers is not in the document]
levkk: Also, we should support "undo".

pgml-sdks/python/pgml/pgml/collection.py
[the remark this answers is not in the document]
montanalow: Yes, this is a perfect example of needing both a local and reference to the same "object" in a single context, and being forced to disambiguate one of them with an `_id` suffix. The confusing ones in this line are `document`, and `splitter`. `chunk` and `chunk_id` are clear and easily distinguished, even though `chunk_id` does not in fact refer to a foreign key in a `chunks` table, it's clearly marked as a reference.

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: In the situation you describe as in any other, we leave it to the author to apply a consistent standard and to judge when disambiguation is necessary. The Railsness here is in the insistence not that a consistent pattern followed -- in this schema, for example, `document` is always an ID -- but rather in insisting that it resemble the schema Rails would have generated. I'm not sure why people with exposure to that platform try to write all SQL that way.

pgml-sdks/python/pgml/pgml/collection.py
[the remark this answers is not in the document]
solidsnack: Or to use it in neither.

pgml-sdks/python/pgml/pgml/database.py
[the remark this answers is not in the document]
montanalow: that's what I meant by "restore"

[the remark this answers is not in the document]
solidsnack: Well, it would be better to leave the schema as is and simply mark it archived in the `collections` table. That makes it easier to query for the condition where it can be deleted: if the `archived` flag is set.

pgml-sdks/python/pgml/pgml/collection.py
montanalow: We write code for others to read. Code should consistently follow the conventions established in the project. The convention has already been established, and there is no need to revisit that convention, when the suggestion obviously confused multiple team members in multiple places due to not just inconsistency with the greater project scope, but inconsistency within the necessary layers to achieve a solution, and also inconsistency within a single table.

solidsnack:
> ...although it would be helpful to define it in a separate .sql file which can then be executed when the SDK is used for the first time.

This is would probably improve things. It can't be a single file, but it could be a few parameterized files -- one for setting up the namespace and main table for a collection, one for setting up each embedding, and maybe some others for the more elaborate queries (as Montana has indicated, some of these queries would benefit from CTEs).

There are some tools in Python for managing queries this way, one of which I wrote ([`query-selector`](https://pypi.org/project/query-selector/), see example [with DDL](https://github.com/solidsnack/query-selector/blob/master/query_selector/example.sql)). It was not a lot of code; maybe 200 lines. I'm looking for some other ones that were updated more recently.

pgml-sdks/python/pgml/pgml/database.py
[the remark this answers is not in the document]
montanalow: That does not guarantee people aren't still using the table, if they are ignoring the flag by using poorly written homegrown queries, which is a potentially desirable outcome for projects that outgrow simple SDK designs, and what to leverage the full expressive power of SQL.

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: It could be called `table_name`, `data_table`, `tab`, &c -- there is nothing wrong with that.

An `oid` is any object ID -- in `pg_class`, the ID of the table is in fact called `oid`. An `oid` the refers to an actual relation can be casted to `regclass`, otherwise not. Since we're dynamically adding and removing tables, it pays to  be conversant with how the catalog describes these things.

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: I don't think there is any other situation, outside of Rails developers talking about schemas, where I've regularly seen people insist that data structures always have fields named according to this rule where timestamps are to end in `_at` and references are to end in `_id`. It doesn't seem to cause confusion with any other programming languages when data structures are not named that way...

pgml-sdks/python/pgml/pgml/database.py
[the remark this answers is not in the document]
solidsnack: In order to access the collection, they look up the collection with the SDK. The SDK looks in the `collections` table for the namespace. The SDK should return an error if the collection is archived.

If they don't look up the namespace with the SDK, how are they going to get the table? People can of course write queries that use the table whether the namespace is moved to `...archived` or not. However, if we have a query builder, that is going to get the namespace and ultimately table for them by first looking in the `collections` table, and the `archived` flag is there.

pgml-sdks/python/pgml/pgml/collection.py
montanalow: Repeatedly using "Rails" as a pejorative isn't productive, or really accurate given it's success as an application development framework, and doesn't really demonstrate a very broad experience with either the Data Mapper or Active Record development patterns commonly used by Object Relational Mappings to deal with the natural impedance mismatch between a relational database schema, and objects in memory, where foreign key references become pointers in memory, to objects in memory, but the foreign key must also always be identifiable in the same context, meaning the case where you only have a reference and not an object is so infrequent that it's not worth dropping the 3 letter suffix and risk causing confusion, particularly in dynamically typed languages where such ambiguity won't be caught at compile time. 

You'll note, the language being used for this does not offer a compiler to check that `document` is in fact an integer, and this is neither Ruby nor Rails, and yet this issue will likely repeat across Javascript and PHP as well. The most common languages for our potential adopters. So not only am I thinking about our internal team, I'm thinking about our customers as well.

pgml-sdks/python/pgml/pgml/collection.py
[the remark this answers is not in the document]
solidsnack: Let's look at the schema in full:

```
CREATE TABLE IF NOT EXISTS %s (
    id          serial8 PRIMARY KEY,
    created_at  timestamptz NOT NULL DEFAULT now(),
    document    int8 NOT NULL REFERENCES
                     ON DELETE CASCADE
                     ON UPDATE CASCADE
                     DEFERRABLE INITIALLY DEFERRED,
    splitter    int8 NOT NULL REFERENCES %s
                     ON DELETE CASCADE
                     ON UPDATE CASCADE
                     DEFERRABLE INITIALLY DEFERRED,
    chunk_id    int8 NOT NULL,
    chunk       text NOT NULL
);
```

What is `chunk_id` actually referencing? There are some global chunks somewhere? Or is it the ordering of the chunk within the document? If it is the ordering, it should be called something else, like `chunk_index` or `position`.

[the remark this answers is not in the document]
santiatpml: `chunk_id` is the ordering of the chunk within the document. `chunk_index` sounds good.

solidsnack (commented):
> ...although it would be helpful to define it in a separate .sql file which can then be executed when the SDK is used for the first time.

As Lev has mentioned, this would improve things -- it would be much easier to follow the schema if it was separated out from the Python code.

It can't be a single file, but it could be a few parameterized files -- one for setting up the namespace and main table for a collection, one for setting up each embedding, and maybe some others for the more elaborate queries (as Montana has indicated, some of these queries would benefit from CTEs).

There are some tools in Python for managing queries this way, one of which I wrote ([`query-selector`](https://pypi.org/project/query-selector/), see example [with DDL](https://github.com/solidsnack/query-selector/blob/master/query_selector/example.sql)). It was not a lot of code; maybe 200 lines. I'm looking for some other ones that were updated more recently.

solidsnack (changes requested):
> ...although it would be helpful to define it in a separate .sql file which can then be executed when the SDK is used for the first time.

As Lev has mentioned, this would improve things -- it would be much easier to follow the schema if it was separated out from the Python code.

It can't be a single file, but it could be a few parameterized files -- one for setting up the namespace and main table for a collection, one for setting up each embedding, and maybe some others for the more elaborate queries (as Montana has indicated, some of these queries would benefit from CTEs).

There are some tools in Python for managing queries this way, one of which I wrote ([`query-selector`](https://pypi.org/project/query-selector/), see example [with DDL](https://github.com/solidsnack/query-selector/blob/master/query_selector/example.sql)). It was not a lot of code; maybe 200 lines. I'm looking for some other ones that were updated more recently.

solidsnack:
SQL Alchemy has something that may work: https://docs.sqlalchemy.org/en/14/core/tutorial.html#using-textual-sql

Alternatively, we could just use Jinja templates, escape carefully, and have one file for each statement/query that needs to be separated out. It's not that bad.

solidsnack (approved):

pgml-sdks/python/pgml/pgml/collection.py
solidsnack: Here, "Rails" isn't a pejorative -- I am simply referring to the community of practice where I've seen the convention.

There are people who don't write tables following the Rails convention where (a) the primary key is always called `id` and the foreign key is always called `<table>_id`. One clue to this is the `USING` keyword for joins -- that actually requires the primary key and the foreign key to have the same name. In our case, that would mean writing schemas like:

```sql
CREATE TABLE IF NOT EXISTS documents_... (
    document    serial8 PRIMARY KEY,
    created_at  timestamptz NOT NULL DEFAULT now(),
    user_uuid   uuid NOT NULL,
    metadata    jsonb NOT NULL DEFAULT '{}',
    text        text NOT NULL,
    UNIQUE (user_uuid)
);

CREATE TABLE IF NOT EXISTS chunks_... (
    chunk       serial8 PRIMARY KEY,
    created_at  timestamptz NOT NULL DEFAULT now(),
    document    int8 NOT NULL REFERENCES documents_...
                     ON DELETE CASCADE
                     ON UPDATE CASCADE
                     DEFERRABLE INITIALLY DEFERRED,
    splitter    int8 NOT NULL REFERENCES splitter_...
                     ON DELETE CASCADE
                     ON UPDATE CASCADE
                     DEFERRABLE INITIALLY DEFERRED,
    position    int8 NOT NULL,
    content     text NOT NULL
);
```

It's actually pretty nice to write `JOIN`s with `USING`; though it is not how Rails schemas are meant to work.

Rails has a broad variety of patterns that it follows; the `_id` and `_at` convention is just one of them. The Rails patterns seem to be helpful in some contexts. When one is writing a lot of templated SQL against a specific database and generating tables on the fly, one is rather outside of the context in which Rails conventions evolved.

montanalow: Natural joins with `USING` is generally useless, because it uses all columns in common, not just the key. In this case `created_at` would be shared but would not be the same, so natural joins won't work. 

```Finally, NATURAL is a shorthand form of USING: it forms a USING list consisting of all column names that appear in both input tables. As with USING, these columns appear only once in the output table. If there are no common column names, NATURAL JOIN behaves like JOIN ... ON TRUE, producing a cross-product join.

Note
USING is reasonably safe from column changes in the joined relations since only the listed columns are combined. NATURAL is considerably more risky since any schema changes to either relation that cause a new matching column name to be present will cause the join to combine that new column as well.
```

montanalow: If you're still not convinced, there is the `user_uuid` in your example, which is yet another instance.

montanalow: But similarly, if you really want to use `USING`, and not natural joins, it works just as well to call them both `document_id`, and you can keep that feature of SQL.

solidsnack: The `USING` clause goes in the same place as `ON`; it's not related to `NATURAL` and doesn't have the limitation you're describing:

```sql
   SELECT document,
          sum(octet_length(chunks.content)) - octet_length(documents.text)
           AS overage
     FROM documents JOIN chunks USING (document)
 GROUP BY document;
```

solidsnack: Calling them both `document_id` would also be a break with Rails conventions, though...

montanalow: You seem confused about the importance of Rails conventions, and to be missing the point that this schema is not an island unto itself, nor is this SDK. The important aspect is that a database and schema in this context are useless without an application layer of logic on top. That application layer will have concerns with ORM that will be simpler to maintain if the database follows a convention of using `_id` prefixes. The reason you see the pattern replicated so widely, is because it's so frequently eases the mental burden of application developers. DBA's or people working only inside the database may not understand these concepts or concerns, but that doesn't mean they don't exist.

solidsnack: In this case, there's not going to be lot of overlap with traditional ORM usage, since the tables are actually created at runtime, binned into schemas, &c; they have to be looked up with the API objects provided. It's really a different pattern than what ORM is for.

I'm just not sure the Rails conventions are replicated that widely or how to get a handle on that objectively -- but another consideration is that Rails conventions are for supporting a style of database usage that is quite different.

montanalow:
You have fixated on rails, without asking about the larger picture.

montanalow (approved):
🥳

PR #644, opened 2023-05-22
Notebook query elapsed time

SilasMarvin:
Added support for dashboard notebooks showing server side time elapsed for queries

### DIFF
--- a/pgml-dashboard/templates/cell.html
+++ b/pgml-dashboard/templates/cell.html
@@ -75,14 +75,7 @@
       <div class="notebook-duration">
         <div class="markdown-body">
           <div class="flex flex-row-reverse">
-            <code>Time: <%- 
-              match cell.execution_time.unwrap().microseconds {
-                microseconds @ 1000000.. => format!("{}s", (microseconds as f64) / 1000000.),
-                microseconds @ 1000..=999999 => format!("{}ms", (microseconds as f64) / 1000.),
-                microseconds @ 0..=999 => format!("{}μs", microseconds),
-                _ => "0".to_string(),
-              }
-            %></code>
+            <code>Time: <%= crate::utils::format_microseconds(cell.execution_time.unwrap().microseconds as f64) %></code>
           </div>
         </div>
       </div>
--- a/pgml-dashboard/templates/cell.html
+++ b/pgml-dashboard/templates/cell.html
@@ -75,7 +75,8 @@
       <div class="notebook-duration">
         <div class="markdown-body">
           <div class="flex flex-row-reverse">
-            <code>TODO</code>
+            <code>Time: <%- (cell.execution_time.unwrap().microseconds as f64) / 1000.0 %>ms</code>
+            <!-- <code>Time: <%- 1 / 1000%>ms</code> -->
           </div>
         </div>
       </div>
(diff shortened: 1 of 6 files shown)


pgml-dashboard/templates/cell.html
```diff
@@ -75,7 +75,14 @@
       <div class="notebook-duration">
         <div class="markdown-body">
           <div class="flex flex-row-reverse">
-            <code>TODO</code>
+            <code>Time: <%-
```
montanalow, line 78: Did you consider moving this inside the `sql.html` `<div class="markdown-body">`? I think that might be better to track each statement independently, but I only thought it through because of the warning about an unused variable in the template.

Otherwise, I'd move it to the cell template that uses it.

SilasMarvin: Sorry I totally missed the unused variable error. I switched back to stable from nightly and can see it now.

That was my original plan, but then I saw in the `models.rs` the `Cell` struct already had an `execution_time` attribute and the `cell.html` had a TODO section for it. 

I just pushed a commit that also shows the timing for each sql statement independently. Let me know what you think! Do you want me to remove the rendering of the `Cell execution_time` in the `cell.html`. If so, should I completely remove the `execution_time` attribute for the `Cell` struct and update the migrations to not include that column?

pgml-dashboard/templates/sql.html
```diff
@@ -24,6 +24,15 @@
       <% } %>
     </tbody>
   </table>
+  <code>
```
montanalow, line 27: It would be nice to hide this block if there is only a single statement, so that we don't double display the timing in that case.

montanalow (commented):
This is looking pretty good to me, with a couple cleanups.

pgml-dashboard/templates/cell.html
```diff
@@ -75,7 +75,14 @@
       <div class="notebook-duration">
         <div class="markdown-body">
           <div class="flex flex-row-reverse">
-            <code>TODO</code>
+            <code>Time: <%- 
+              match cell.execution_time.unwrap().microseconds {
+                microseconds @ 1000000.. => format!("{}s", (microseconds as f64) / 1000000.),
+                microseconds @ 1000..=999999 => format!("{}ms", (microseconds as f64) / 1000.),
+                microseconds @ 0..=999 => format!("{}μs", microseconds),
+                _ => "0".to_string(),
```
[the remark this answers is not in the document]
SilasMarvin: If I remove the bottom one I get a compiler error: `error[E0004]: non-exhaustive patterns: i64::MIN..=-1_i64 not covered`.

I'm using rust 1.69.0

pgml-dashboard/templates/sql.html
SilasMarvin: Added a utils.rs module with a function for doing this. I made it take an `f64`, as we convert it to that regardless of whether we are working with a `std::time::Duration` or `postgres Interval`.

pgml-dashboard/templates/sql.html
```diff
@@ -24,6 +24,15 @@
...
+    Time: 
+    <%- match execution_duration.as_micros() {
```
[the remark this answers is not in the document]
SilasMarvin: Added a utils.rs module with a function for doing this. I made it take an f64, as we convert it to that regardless of whether we are working with a std::time::Duration or postgres Interval.

pgml-dashboard/templates/sql.html
SilasMarvin: Done!

montanalow (approved):

PR #655, opened 2023-05-24
Add LangChain splitters

levkk:


pgml-extension/src/bindings/chunking.py
```diff
@@ -0,0 +1,29 @@
+from langchain.text_splitter import (
```
montanalow, line 1: I'd rename this to langchain, since other bindings refer to the library dependency rather than the functionality.

levkk: Yup

PR #659, opened 2023-05-29
Added random_state as an available hyper parameter for xgboost models

SilasMarvin:
I followed the standard found throughout the repo and just called `unwrap` on line 166. This does have the potential to cause somewhat confusing error messages if they have malformed arguments. I would be happy to go through and replace the `unwrap()` calls with `expect(SOME HELPFUL ERROR MESSAGE)` throughout pgml-extension if that would be helpful?


montanalow (approved):

montanalow:
You're right. `expect()` is preferable to `unwrap()`, but rather than trying to convert all existing cases, I think it'd be better to try to move most functions that currently `unwrap` to instead return an `anyhow::Result`, and add `context()` to their calls for better nested function call clarity.

I think the current `transformer` bindings for `transform` and `embed` are the worst offenders with the mostly likely failure cases, and the scikit algorithm's are probably second worst on that front.

PR #682, opened 2023-06-04
Adding filtering based on metadata using kwargs

jsaied99:
Improving the vector search by adding filtering based on the additional kwargs. Adding from [#663](https://github.com/postgresml/postgresml/issues/663). 

You can add additional metadata filters by passing them in the `vector_search` call. For example, if you want to filter by title:

`results = collection.vector_search(query, top_k=5, title='Beyonce')` or by page_number
`results = collection.vector_search(query, top_k=5, page_number=13)`


montanalow (approved):
This adds flat single level filters, which I think is good enough to improve utility, and I think we can probably save further filtering operations and optimizations for follow ups, like GIN indexes.

PR #683, opened 2023-06-04
Adding text_search using similarity from pg_trim and tests for text_search and metadata filtering

jsaied99:
text_search functionality by doing `collection.text_search(query, top_k=5)`

### DIFF
--- a/pgml-sdks/python/pgml/README.md
+++ b/pgml-sdks/python/pgml/README.md
@@ -282,8 +282,8 @@ LOGLEVEL=INFO python -m unittest tests/test_collection.py
 
 ### Roadmap
 
-- Enable filters on document metadata in `vector_search`. [Issue](https://github.com/postgresml/postgresml/issues/663)
-- `text_search` functionality on documents using Postgres text search. [Issue](https://github.com/postgresml/postgresml/issues/664)
+- ~~Enable filters on document metadata in `vector_search`. [Issue](https://github.com/postgresml/postgresml/issues/663)~~
+- ~~`text_search` functionality on documents using Postgres text search. [Issue](https://github.com/postgresml/postgresml/issues/664)~~
 - `hybrid_search` functionality that does a combination of `vector_search` and `text_search` in an order specified by the user. [Issue](https://github.com/postgresml/postgresml/issues/665)
 - Ability to call and manage OpenAI embeddings for comparison purposes. [Issue](https://github.com/postgresml/postgresml/issues/666)
 - Save `vector_search` history for downstream monitoring of model performance. [Issue](https://github.com/postgresml/postgresml/issues/667)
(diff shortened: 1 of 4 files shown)


montanalow (commented):
I would expect `text_search` to use tsvector and GIN indexes, to implement recall by keywords, which is different than something I might call `fuzzy_search` with `pg_trgm` (which can also use a GIN index). I think it's worth spending some time to think through the workflow to combine vector, keyword and trigram strategies for both recall and ranking steps, and how we can do all of those in a single query with a single unified document result list, that includes multiple scores and possibly re-ranks using a weighted combination (or even a model).

Since we need to build the vector index as a distinct step, I think we should probably require distinct steps to build the keyword and trigram indexes as well, before those kind of queries can be executed efficiently. 

This implies we'll need to implement more of a query builder pattern to combine these sorts of options in an API. i.e. `vector_recall(text, limit)` should return a query builder rather than an immediate query result, and we should be able to chain calls `keyword_recall(text, limit)` and `fuzzy_recall(text, limit)` on the intermediate results. Then we can have a finalizer maybe named `results(limit)` that actually executes the query and returns the data from the db. 

cc @SilasMarvin is also looking into porting this SDK to Rust and Node, so I think it's worth spending an iteration on the API design.

https://www.postgresql.org/docs/current/textsearch.html
https://www.postgresql.org/docs/current/pgtrgm.html#id-1.11.7.44.8

montanalow:
Additionally, we'll probably want to combine your previous PR to add `filter` on the query builder, and additionally the ability to recall by equality matches on the document metadata as well, not just filter.

montanalow:
I think our API needs to consist of composable functions to generate sql fragements, rather than adding a bunch of duplicate-ish functions that support some subsets.

I imagine a pgml.query module with a Query class that has an interface like like:

- `def vector_recall(self, str) -> self`
- `def text_recall(self, str) -> self`
- `def trigram_recall(self, str) -> self`
- `def limit(self, int) -> self`
- `def order(self, str) -> self`
- `def filter(self, dict) -> self`

So I can write code like:

``` 
collection.query().vector_recall('hi mom').limit(10).fetch() -> list[Document]
```
or 
``` 
collection.query().text_recall('hi mom').limit(10).fetch() -> list[Document]
```
or 
``` 
collection.query().vector_recall('hi mom').text_recall('hi mom').filter({"name": "bob"}).limit(10).fetch() -> list[Document]
```

Each intermediate result after `collection.query()` is a `Query`, and we can start building other query functions that take queries as  arguments to generate CTE's or subqueries for more advanced use cases. With a query object like this, we can also develop more complex (AND/OR/NOT) predicates as well for filters.

santiatpml:
Pypika provides a [good starting point](https://pypika.readthedocs.io/en/latest/_modules/pypika/queries.html#CreateQueryBuilder) to add custom queries. We can pull some of these in.

PR #685, opened 2023-06-05
Development dockerfile

jsaied99:
Fixing `Dockerfile.local` for local development. To run it you can do:

`docker compose -f docker-compose.local.yml build --build-arg PG_MAJOR_VER=14`
`docker compose -f docker-compose.local.yml  up`


montanalow (approved):
This is a great improvement. I think the next step will be to also create a distinction between local and packaged versions of pgml-dashboard to mirror the extension, and then update our packaging github actions, so we're consistent all the way.

PR #686, opened 2023-06-05
Adding embed_array for getting the embeddings of multiple strings

jsaied99:
Works the same way as pgml.embed but can take an array of inputs. Example:

```sql
SELECT pgml.embed(
    transformer => 'intfloat/e5-small', 
    inputs => ARRAY['Hello', 'World'],
    kwargs => '{"device": "cpu"}'::JSONB
);
```

```sql
SELECT pgml.embed(
    transformer => 'hkunlp/instructor-base', 
    inputs => ARRAY['Hello World', 'I love Rust'],
    kwargs => '{"device": "cpu", "instruction": "Represent the content for retrieving supporting documents:"}'::JSONB
);
```

For instructor, I'm passing the same instruction to each input, we could potentially have a inputs array that is a json and each individual input.

### DIFF
--- a/pgml-extension/src/api.rs
+++ b/pgml-extension/src/api.rs
@@ -573,7 +573,7 @@ pub fn embed_array(
     transformer: &str,
     inputs: Vec<&str>,
     kwargs: default!(JsonB, "'{}'"),
-) -> Vec<f32> {
+) -> Vec<Vec<f32>> {
     crate::bindings::transformers::embed_array(transformer, &inputs, &kwargs.0)
 }
 
--- a/pgml-extension/src/api.rs
+++ b/pgml-extension/src/api.rs
@@ -569,12 +569,12 @@ pub fn embed(transformer: &str, text: &str, kwargs: default!(JsonB, "'{}'")) -> 
 }
 
 #[pg_extern(immutable, parallel_safe, name = "embed")]
-pub fn embed_array(
+pub fn embed_batch(
     transformer: &str,
     inputs: default!(Vec<String>, "ARRAY[]::TEXT[]"),
     kwargs: default!(JsonB, "'{}'"),
 ) -> Vec<Vec<f32>> {
-    crate::bindings::transformers::embed_array(transformer, &inputs, &kwargs.0)
+    crate::bindings::transformers::embed_batch(transformer, &inputs, &kwargs.0)
 }
 
 #[pg_extern(immutable, parallel_safe)]
--- a/pgml-extension/src/api.rs
+++ b/pgml-extension/src/api.rs
@@ -569,8 +569,12 @@ pub fn embed(transformer: &str, text: &str, kwargs: default!(JsonB, "'{}'")) -> 
 }
 
 #[pg_extern(immutable, parallel_safe)]
-pub fn embed(transformer: &str, inputs: Vec<&str>, kwargs: default!(JsonB, "'{}'")) -> Vec<f32> {
-    crate::bindings::transformers::embed(transformer, &inputs, &kwargs.0)
+pub fn embed_array(
+    transformer: &str,
+    inputs: Vec<&str>,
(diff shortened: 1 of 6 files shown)


pgml-extension/src/bindings/transformers.py
```diff
@@ -110,22 +110,38 @@ def transform(task, args, inputs):
     return json.dumps(pipe(inputs, **args), cls=NumpyJSONEncoder)
 
 
-def embed(transformer, text, kwargs):
+def embed(transformer, inputs, kwargs):
+    
+    try:
+        inputs = json.loads(inputs)
+    except json.decoder.JSONDecodeError:
```
montanalow, line 117: What is the known case that this is handling?

pgml-extension/src/bindings/transformers.py
```diff
@@ -110,22 +110,38 @@ def transform(task, args, inputs):
...
+    
     if instructor:
         klass = INSTRUCTOR
-        text = [[kwargs.pop("instruction"), text]]
+        if isinstance(inputs, str):
+            inputs = [[kwargs.pop("instruction"), inputs]]
+            
+        else:
+            texts_with_instructions = []
+            instruction = kwargs.pop("instruction")
```
montanalow, line 131: Seems like it may be common to have multiple instructions with multiple inputs? Hmm, in that case I'm not sure we have a nice way to structure the args... we can leave that for some future work.

pgml-extension/src/bindings/transformers.py
```diff
@@ -110,22 +110,38 @@ def transform(task, args, inputs):
...
 
     if transformer not in __cache_sentence_transformer_by_name:
         __cache_sentence_transformer_by_name[transformer] = klass(transformer)
     model = __cache_sentence_transformer_by_name[transformer]
 
-    result = model.encode(text, **kwargs)
-    if instructor:
+    result = model.encode(inputs, **kwargs)
+    if instructor and len(result) == 1:
         result = result[0]
```
montanalow, line 145: I think we also need to handle the multi instructor case, right? Or does instructor always just return an appropriate single dimension array?

pgml-extension/src/api.rs
```diff
@@ -568,6 +568,15 @@ pub fn embed(transformer: &str, text: &str, kwargs: default!(JsonB, "'{}'")) ->
     crate::bindings::transformers::embed(transformer, text, &kwargs.0)
 }
 
+#[pg_extern(immutable, parallel_safe, name = "embed")]
+pub fn embed_batch(
+    transformer: &str,
+    inputs: default!(Vec<String>, "ARRAY[]::TEXT[]"),
+    kwargs: default!(JsonB, "'{}'"),
+) -> Vec<Vec<f32>> {
+    crate::bindings::transformers::embed_batch(transformer, &inputs, &kwargs.0)
```
montanalow, line 577: I would make `crate::bindings::transformers::embed` take a list of inputs always, and modify `pub fn embed` to pass a slice in, similar to `generate` and `generate_batch`.

jsaied99: Sure, I'll change this. Makes sense

pgml-extension/src/bindings/transformers.py
jsaied99: Okay I'll leave it as it is.

pgml-extension/src/bindings/transformers.py
jsaied99: Without passing any extra arguments, I couldn't think of a way of knowing wether inputs was a string or a JSON string. Thought the simplest way was trying to convert it into a python object.

pgml-extension/src/api.rs
jsaied99: I guess if we do pass an input array always, I could get rid of the try catch since it's always an array of strings.

montanalow (approved):

PR #689, opened 2023-06-05
metadata and generic filters in vector search

santiatpml:
I have added two filters to vector_search: `metadata_filter` uses basic containment operator  `@>` and `generic_filter` takes pretty much anything that the user wants to filter on `documents` table using SQL syntax. 

`generic_filter` should give us quite a bit of coverage including timestamps and nested metadata.

Added three tests.

### DIFF
--- a/pgml-sdks/python/pgml/pgml/collection.py
+++ b/pgml-sdks/python/pgml/pgml/collection.py
@@ -298,18 +298,23 @@ class Collection:
                 )
                 continue
 
+            metadata = document
+
             _uuid = ""
             if id_key not in list(document.keys()):
                 log.info("id key is not present.. hashing")
-                source_uuid = hashlib.md5(text.encode("utf-8")).hexdigest()
+                source_uuid = hashlib.md5(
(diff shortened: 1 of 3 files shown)


pgml-sdks/python/pgml/pgml/collection.py
```diff
@@ -784,9 +780,20 @@ def vector_search(
             top_k=top_k,
             chunks_table=self.chunks_table,
             documents_table=self.documents_table,
-            metadata_filter=metadata_filter,
         )
 
+        if metadata_filter:
+            cte_select_statement += (
+                " AND documents.metadata @> {metadata_filter}".format(
```
montanalow, line 787: This is nice. We should also add a GIN index to `documents.metadata`.

montanalow (approved):
I think there is a quickly growing need to move to a builder pattern, to prevent having to duplicate this functionality for example on a text_search or fuzzy_search function. We need a better API that more cleanly encapsulates recall, ranking and filtering, reranking and refiltering, and finally passing as an input to another SQL function or recalling from the database to the application.

PR #696, opened 2023-06-06
Update sql interface for 2.5.0

montanalow:


pgml-extension/sql/pgml--2.5.0--2.5.1.sql
```diff
@@ -0,0 +1,24 @@
+-- src/api.rs:575
+-- pgml::api::embed
+CREATE  FUNCTION pgml."embed"(
```
levkk, line 3: Or replace.

Issue #715, opened 2023-06-09
Incorrect operator used in pgvector example for the query to use the index


amirGhasemi:
In the pgvector example mentioned in [embedding](https://postgresml.org/docs/guides/transformers/embeddings) section of the documentation, a wrong operator is used for the query to take advantage of the index. The `<->` operator is used for L2 distance while `<=>` is the corect operator for cosine distance. The created index uses `pgvector.vector_cosine_ops` operator class. So the correct operator to use should be `<=>`.
Similiarly, for the query to use the index, one could change the operator class to `pgvector.vector_l2_op`.

For more information regarding pgvector, see the following sections of the pgvector README.md:
[Indexing](https://github.com/pgvector/pgvector/tree/master#indexing)
[Vector Operators](https://github.com/pgvector/pgvector/blob/master/README.md#vector-operators)

amirGhasemi:
There also following problems:

- The index is created on the wrong table(`tweet_embeddings_2` instead of `items`).
- The index is created after the query is executed.
- The `tweet_embeddings_2` table is not created in the same documetation.
- There is an extra identical example which only differs from the previous one in table name(`tweet_embeddings_2 ` instead of `tweet_embeddings`) and model name(`sentence-transformers/all-MiniLM-L6-v2` instead of `distilbert-base-uncased`). The extra example does not add any new information to the documentation and it disrupts the flow of the guide. I think it is better to remove it altogether.

levkk:
Thank you for reporting this! A PR is welcome, if you have a moment to help us with the fix.

amirGhasemi:
All of the mnetioned suggestion are done in this [PR](https://github.com/postgresml/postgresml/pull/716).

PR #716, opened 2023-06-09
Update embeddings.md to use cosine distance in pgvector example

amirGhasemi:
#715
Used cosine distance operator in query to take advantage of the index.


levkk (approved):
Good find!

PR #720, opened 2023-06-09
Dan guards fairings

chillenberger:
- Make current user a global state and remove user info from cluster. 
- Get user from global state and no longer use a guard to return user info.
- Make cluster an option so we can distinguish between no user and default local user. 
- If no cluster is set in state, throw 500 error if user tries to navigate to dashboard.  
- Remove dashboard link if no cluster in global state. 
- Remove footer links (login, terms of service, privacy policy, create an account) when dashboard is run locally. 

https://github.com/postgresml/postgresml/assets/39170265/2746df17-b8f6-4ee2-af56-86f615ab7efb


pgml-dashboard/src/lib.rs
```diff
@@ -119,6 +118,44 @@ impl Clusters {
     }
 }
 
+// Globally shared state in memory.
+
+#[derive(Debug, Clone)]
+pub struct CurrentUser {
```
levkk, line 124: I'm not sure what this does. Is this meant to store the current user in the HTTP request?

chillenberger: It is a global state to hold the currently signed in user and the clusters they have visible to them.  It was inside cluster context before, but you can have a user without a chosen cluster.

levkk: I understand now. This is not going to work as is, I think. This global is global for everyone using the dashboard. In our cloud, this can be many different people. In this case, person A can land on the dashboard and we set their user as `test@test.com`. Then person B lands on the dashboard and we set this global user, in the same memory space, to `test2@test.com` and this overrides the same setting for person A.

What we do in the `Clusters` context is we set the current cluster in a `HashMap` with the `cluster_id` <-> `Cluster` object mapping. So when a request comes in, we fetch the request-specific object from the HashMap using the `cluster_id` key.

The global in that case stores _everyone_'s data and we retrieve the data specific to the user from that global.

pgml-dashboard/src/lib.rs
```diff
@@ -626,3 +660,93 @@ pub fn routes() -> Vec<Route> {
...
+                new_visible_clusters.get("1"), 
+                new_visible_clusters.get("2")
+            )
+        )
+    }
+
+    // It should reset current user and visible clusters to none. 
+    #[test]
+    fn test_reset_user_to_default() {
+        let test_user = CurrentUser::new();
```
levkk, line 720: This is single-threaded, single-request simulation. I think you'll be able to validate that the current implementation doesn't work if you spawn 1000 threads and make them set and fetch the current user very quickly in succession.

pgml-dashboard/src/lib.rs
```diff
@@ -119,6 +118,44 @@ impl Clusters {
...
+
+    pub fn get_user(&self) -> Option<models::User> {
+        self.user.lock().clone()
+    }
+
+    pub fn set_visible_clusters(&self, visible_clusters: HashMap<String, String>) {
+        *self.visible_clusters.lock() = Some(visible_clusters);
+    }
+
+    pub fn get_visible_clusters(&self) -> Option<HashMap<String, String>> {
```
levkk, line 142: I think we have a `Cluster` object we make public. We can use it here to benefit from some type safety.

pgml-dashboard/src/lib.rs
```diff
@@ -626,3 +660,93 @@ pub fn routes() -> Vec<Route> {
...
+            new_user_data.email,
+        ));
+    }
+
+    // It should be possible to set and get visible clusters. 
+    #[test]
+    fn test_set_get_visible_clusters() {
+        let test_user = CurrentUser::new();
+        let new_visible_clusters: HashMap<String, String> = HashMap::from([
+            ("1".to_string(), "123456789".to_string()),
```
levkk, line 702: The cluster IDs are actually `i64` (integer).

pgml-dashboard/src/lib.rs
chillenberger: ok I see. I think a lot of this code is still usable.  I could probably change CurrentUser to take a HashMap, more like Clusters, and adjust the getter and setter methods.

PR #721, opened 2023-06-09
Working Rust sdk

SilasMarvin:


pgml-sdks/rust/pgml/src/collection.rs
```diff
@@ -0,0 +1,577 @@
...
+    get_text_splitters,
+    generate_chunks,
+    register_model,
+    get_models,
+    generate_embeddings,
+    vector_search,
+    get_name
+)]
+impl Collection {
+    pub async fn new(name: String, pool: Arc<PgPool>) -> anyhow::Result<Self> {
```
levkk, line 43: `PgPool` is already `Sync` and `Send`, you don't need an `Arc`.

PR #722, opened 2023-06-10
Adding a way to clear GPU memory 

jsaied99:
Adding a way of clearing GPU memory by doing `SELECT pgml.clear_gpu_cache();`. One can also pass a `memory_usage` parameter which will add a filter to whether or not to clear GPU memory if utilization is greater than the passed value. For instance, if you want to clear only if GPU memory utilization is > 50%, you can do the following:

` select pgml.clear_gpu_cache(memory_usage => 50);`

I also wanted to implement the general clear_cache, but it requires a bit more, and I think clearing GPU should be okay for the first pass.


montanalow (changes requested):
This looks good to me. 2 nits. 

1) We should start including all this wonderful PR example/documentation as crate docs over API functions. i.e. /// blah blah blah
2) I think we should use a f32 for the memory limit argument since a fraction between [0:1]

jsaied99:
@montanalow , 

1. `torch.cuda.memory_usage` [Torch Documentation](https://pytorch.org/docs/stable/generated/torch.cuda.memory_usage.html) returns a int. 
2. Makes sense to me, I'll add docs starting for this PR. Not that familiar with the format for the crate docs, but shouldn't be an issue.

montanalow:
I appreciate you beginning the great documentation crusade with this function. For the argument type, all of our other APIs follow more scikit-esque floating point numbers. Torch is the most common backend, maybe the only one that matters these days, but I think floating point is still the better API for relative percentages, and I think consistency is generally important.

jsaied99:
Got it. I agree with consistency. No worries. I will change it to `f32` and make the proper adjustments in the code.

montanalow (approved):

PR #730, opened 2023-06-13
Generate python stub files in proc macros

SilasMarvin:


montanalow:
One thing we should eventually do, is seperate the CI test suites on github for the SDK and the extension

PR #731, opened 2023-06-13
xformers installation fix

santiatpml:


pgml-extension/requirements.txt
```diff
@@ -18,7 +18,6 @@ torchaudio==2.0.2
 torchvision==0.15.2
 tqdm==4.65.0
 transformers==4.29.2
-xformers==0.0.20
```
levkk, line 21: We use this file to install dependencies elsewhere, e.g. in our production cloud. Would it be possible to add another file, perhaps `requirements-xformers.txt` to install that special package?

levkk: I think you can do:

_requirements.txt_
```
xformers==0.0.20 --no-dependencies
```

which is then installable with 

```
pip install -r requirements-xformers.txt
```

levkk: I'm afraid that if we put this package only in the dockerfile, we'll forget to add it elsewhere where it's needed.

santiatpml: `--no-dependencies` has to be outside requirements.txt file.
`pip install -r requirements-xformers.txt --no-dependencies`

PR #733, opened 2023-06-13
Docker fix of the day

levkk:
### Bugs

- Fix dashboard not starting up because of missing dependencies: npm, nodejs, sass.
- Bust Docker cache on `pgml-exension` build when PostgresML version changes
- Remove compilation warning for `pgml-dashboard` (extra semicolumn).
- Don't fail `setup_examples.sql` if things already exist. We have a volume now, they will fail every time the container starts up otherwise.

### Features

- Move around things in `pgml-extension/Dockerfile` to benefit more from Docker's layers and build things faster.

### DIFF
--- a/pgml-extension/examples/binary_classification.sql
+++ b/pgml-extension/examples/binary_classification.sql
@@ -1,5 +1,5 @@
 -- Exit on error (psql)
-\set ON_ERROR_STOP true
+-- \set ON_ERROR_STOP true
 \timing on
 
 SELECT pgml.load_dataset('breast_cancer');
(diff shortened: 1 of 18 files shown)


pgml-extension/examples/binary_classification.sql
```diff
@@ -1,5 +1,5 @@
 -- Exit on error (psql)
-\set ON_ERROR_STOP true
+-- \set ON_ERROR_STOP true
```
montanalow, line 2: I think this is going to cause a problem later when we miss breakage. These are intended to be part of CI.

levkk: https://github.com/postgresml/postgresml/blob/master/pgml-extension/tests/test.sql#L9

PR #748, opened 2023-06-16
GGML and GPTQ compatibility

### LINKED ISSUE #717: any plans to add support for llama.cpp?
this would enable to leverage large language models.

montanalow:
Fix #717 

GPT 2 comparison:
~ggml version is 4x faster~, I'm no longer able to reproduce a significant speed difference in just gpt2, but that may be because it all fits in GPU either way.

```
SELECT pgml.transform(
    task => '{
      "task": "text-generation",
      "model": "gpt2"
    }'::JSONB,
    inputs => ARRAY[
        'Once upon a time,',
        'It was the best of times'
    ],
    args => '{"max_new_tokens": 32}'::JSONB
);

                                                                                                                                                                                   transform

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 [[{"generated_text": "Once upon a time, they felt the feeling. Now, it would not be the first time that this was happening with them, nor would it be the first time that it had occurred"}]
, [{"generated_text": "It was the best of times,\" the actor replied.\n\nSue Pearce, a New York theater actress who plays the mother of Sam (Cameron Diaz) on \"House of Cards"}]]
(1 row)

Time: 458.381 ms
```

```
SELECT pgml.transform(
    task => '{
      "task": "text-generation",
      "model": "marella/gpt-2-ggml"
    }'::JSONB,
    inputs => ARRAY[
        'Once upon a time,',
        'It was the best of times'
    ],
    args => '{"max_new_tokens": 32}'::JSONB
);

                                                                                                                     transform

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
------------------------------------------------------
 [" I was not only able to make the best of my situation but also have some fun with it.\n\nI've been playing games for years now and even", ". I didn't have to go through that, but it's no
t like you can do anything else with your life.\"\n"]
(1 row)

Time: 455.720 ms
```

```
SELECT pgml.transform(
    task => '{
      "task": "text-generation",
      "model": "mlabonne/gpt2-GPTQ-4bit"
    }'::JSONB,
    inputs => ARRAY[
        'Once upon a time,',
        'It was the best of times'
    ],
    args => '{"max_new_tokens": 32}'::JSONB
);

                                                                                                                                              transform

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------
 ["Once upon a time, the world was a place of great beauty and great danger. The world was a place of great danger. The world was a place of great danger. The world", "It was the best of ti
mes.\n\n\"I'm not going to be a part of that team,\" he said. \"I'm going to be a part of the team. I"]
(1 row)

Time: 577.400 ms
```

### DIFF
--- a/pgml-extension/requirements.txt
+++ b/pgml-extension/requirements.txt
@@ -1,4 +1,5 @@
 accelerate==0.19.0
+auto-gptq==0.2.2
 ctransformers==0.2.8
 datasets==2.12.0
 deepspeed==0.9.2
--- a/pgml-extension/requirements.txt
+++ b/pgml-extension/requirements.txt
@@ -1,4 +1,5 @@
 accelerate==0.19.0
+ctransformers==0.2.8
 datasets==2.12.0
 deepspeed==0.9.2
 huggingface-hub==0.14.1
--- a/pgml-extension/src/bindings/transformers.py
+++ b/pgml-extension/src/bindings/transformers.py
@@ -104,7 +104,7 @@ def transform(task, args, inputs):
 
     if type(pipe) is ctransformers.llm.LLM:
         # ctransformers don't support batch inference
-        results = pipe(inputs[0], **args)
+        results = [pipe(inputs[0], **args)]
     else:
         if pipe.task == "question-answering":
             inputs = [orjson.loads(input) for input in inputs]
--- a/pgml-extension/src/bindings/transformers.py
+++ b/pgml-extension/src/bindings/transformers.py
@@ -3,6 +3,7 @@ import os
 import shutil
 import time
 
+import ctransformers
 import datasets
 from InstructorEmbedding import INSTRUCTOR
 import numpy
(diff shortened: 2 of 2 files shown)


levkk:
Huh that's interesting. Would be cool to try llama too, if it's on HF. I wonder if the Rust interface is faster? Probably not significantly. I think it would be "cooler" to implement the `llm` crate eventually just so we can run some of these models without Python.

Tostino:
The quality does often degrade slightly with quantization for the same model, but you can fit a larger, more capable model in the same RAM when you are using a quantized model vs full precision. I can fit a 30b parameter model on 1x 3090s, or a 65b parameter on 2x 3090s with either GGML or GPTQ quantization. Those will give vastly better answers than the largest full precision models I can fit on the same hardware.

VGerris:
@Tostino can you explain or point to documentation that shows how to set up multiple GPUs and use them ? Thank you!

PR #762, opened 2023-06-19
v2.6.0

montanalow:
Add a blog post for 2.6.0 highlighting GPTQ and GGML


pgml-extension/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM nvidia/cuda:12.1.1-devel-ubuntu22.04
 LABEL maintainer="team@postgresml.com"
 
 ARG DEBIAN_FRONTEND=noninteractive
-ARG PGML_VERSION=2.5.3
+ARG PGML_VERSION=2.6.0
```
levkk, line 5: This needs to be updated after with release the extension into the wild via our github actions job.

montanalow: Why are we hardcoding this specifically? Why don't we use `apt-get upgrade` instead of installing a fixed version, so that we can avoid this 2 step and release dependency?

levkk: Because Docker caches the apt-get call and users end up running old versions of the extension every time they do docker-compose up.

levkk: Docker requires one to change the line in the Dockerfile to bust the cache.

PR #773, opened 2023-06-21
Replace model.generate call with a pipeline call for GPTQPipeline. 

Tostino:
Replace model.generate call with a pipeline call for GPTQPipeline.  Made the non-quantized path work the same way as GPTQPipeline. The pipeline interface is what we want here. It allows us to pass in args => '{"return_full_text": false}'::JSONB to get similar output as we get for GGML models.

### DIFF
--- a/pgml-extension/src/bindings/transformers.py
+++ b/pgml-extension/src/bindings/transformers.py
@@ -36,6 +36,7 @@ from transformers import (
     GenerationConfig,
     TrainingArguments,
     Trainer,
+    pipeline
 )
 
 __cache_transformer_by_model_id = {}
@@ -82,6 +83,27 @@ def ensure_device(kwargs):
             kwargs["device"] = "cpu"
 
 
+class StandardPipeline(object):
+    def __init__(self, model_name, **task):
+        task.pop("model")
+        self.task = task.pop("task")
+        task.pop("device")
+        self.model = AutoModelForCausalLM.from_pretrained(model_name, **task)
+        self.tokenizer = AutoTokenizer.from_pretrained(model_name)
+
+    def __call__(self, inputs, **kwargs):
+        outputs = []
+        pipe = pipeline(
+            self.task,
+            model=self.model,
+            tokenizer=self.tokenizer,
+            **kwargs
+        )
+        for input in inputs:
+            outputs.append(pipe(input)[0]['generated_text'])
+        return outputs
+
+
 class GPTQPipeline(object):
     def __init__(self, model_name, **task):
         import auto_gptq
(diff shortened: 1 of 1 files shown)


pgml-extension/src/bindings/transformers.py
```diff
@@ -97,10 +119,14 @@ def __init__(self, model_name, **task):
...
+            self.task,
+            model=self.model,
+            tokenizer=self.tokenizer,
+            **kwargs
+        )
         for input in inputs:
-            tokens = self.tokenizer(input, return_tensors="pt").to(self.model.device).input_ids
-            token_ids = self.model.generate(input_ids=tokens, **kwargs)[0]
-            outputs.append(self.tokenizer.decode(token_ids))
+            outputs.append(pipe(input)[0]['generated_text'])
```
montanalow, line 129: The result of the pipe, `generated_text`, doesn't seem to be 100% standardized across models. I think this will cause an error for those cases that return a string instead of a dict.

pgml-extension/src/bindings/transformers.py
```diff
@@ -82,6 +83,27 @@ def ensure_device(kwargs):
...
+    def __init__(self, model_name, **task):
+        task.pop("model")
+        self.task = task.pop("task")
+        task.pop("device")
+        self.model = AutoModelForCausalLM.from_pretrained(model_name, **task)
+        self.tokenizer = AutoTokenizer.from_pretrained(model_name)
+
+    def __call__(self, inputs, **kwargs):
+        outputs = []
+        pipe = pipeline(
```
montanalow, line 96: We don't want to incur the cost of instantiating a new pipeline on every invocation. This is something that should happen during __init__.

Tostino: Don't disagree. I didn't look into if pipeline was an expensive instantiation or really just a wrapper. I put it in the call rather than init to allow different parameters (temp, top_k, etc) per call without having to load a whole new model into ram and cache it.

pgml-extension/src/bindings/transformers.py
Tostino: Do you have an example model that is returning something different so I can make this more robust?

pgml-extension/src/bindings/transformers.py
Tostino: Yup, looks heavy enough to not do that: https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/base.py#L735

pgml-extension/src/bindings/transformers.py
montanalow: I think any transformer that doesn't use the `text-generation` task will have different output format, like `translation`, `summarization` or `text-classification`. I'm not sure if those tasks also have specific pipeline output formats. That'd be interesting to also handle more cleanly, but changing the output format would also be a breaking API change for anyone currently using it as is.

PR #790, opened 2023-06-29
Upgrade to pgrx 0.9.7

workingjubilee:
A critical bug concerning handling arrays with an initial null element was discovered in pgrx versions 0.8.0 to 0.9.6, recommending an upgrade to 0.9.7 as soon as is possible. I was not able to run the necessary commands to update the lockfiles, my apologies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
         run: |
           curl https://sh.rustup.rs -sSf | sh -s -- -y
           source ~/.cargo/env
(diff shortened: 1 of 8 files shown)


montanalow:
Thanks @workingjubilee. Tests pass for me with your changes. What lockfile updates were you prevented from making? We don't typically `cargo update` every time we update a dependency.

workingjubilee:
Fair enough! I was trying to run `cargo update -p pgrx` on the relevant crates, but if it's not a big deal it's not a big deal.

PR #836, opened 2023-07-18
Dan webapp nav redesign

chillenberger:
- Web app top nav mobile and large screens, 
- Web app left nav, dropdown, with collapse, 
- db update and routing

https://github.com/postgresml/postgresml/assets/39170265/6d7e6453-d303-47fb-a305-a5b6894ffe81


pgml-dashboard/src/templates/mod.rs
```diff
@@ -138,70 +141,66 @@ impl<'a> WebAppBase<'a> {
...
+                    // NavLink::new("Notebooks", "/notebooks")
+                    //     .icon("description")
+                    //     .disable(true),
+                    // NavLink::new("Explore", "/explore")
+                    //     .icon("thumbnail_bar")
+                    //     .disable(true),
+                    // NavLink::new("Projects", "/projects")
+                    //     .icon("library_add")
+                    //     .disable(true),
+                    NavLink::new("Status", &format!("/clusters/{}", self.current_cluster.id))
```
levkk, line 171: At some point, we would like to be able to pass these in from our cloud app, because the URLs and names could change, but for now, this will work.

pgml-dashboard/src/utils/url_parser.rs
```diff
@@ -0,0 +1,128 @@
...
+    static ref DOMAIN_REGEX: Regex =
+        Regex::new(r"^https?://((localhost:8000)|(.*\.?postgresml\.org))").unwrap();
+    static ref UUID_REGEX: Regex = Regex::new(".{8}-.{4}-.{4}-.{4}-.{12}").unwrap();
+}
+
+pub struct ClusterUrlParser;
+
+impl ClusterUrlParser {
+    // Returns option containing path. If regex does not match, returns None.
+    pub fn get_path(url: &str) -> Option<String> {
```
levkk, line 14: Rocket allows you to get a path without resorting to a regex: https://docs.rs/rocket/latest/rocket/struct.Request.html#method.uri

pgml-dashboard/templates/components/github_icon.html
```diff
@@ -1,8 +1,8 @@
 <a class="d-flex align-items-center nav-link p-0 border-bottom-0" href="https://github.com/postgresml/postgresml">
-<% if let Ok(stars) = crate::utils::config::github_stars() { %>
-    <span class="badge github-badge"><p>Stars | <%= stars %></p></span>
-<% } %>
-    <svg width="35" height="35" viewBox="0 0 40 39" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <% if show_stars {%>
```
levkk, line 2: The stars should not be hardcoded, we fetch them from Github on every build.

pgml-dashboard/templates/layout/nav/top_web_app.html
```diff
@@ -0,0 +1,171 @@
+<% use crate::templates::components::GithubIcon; %>
+<% use crate::templates::components::PostgresLogo; %>
+<% 
+  let account_management_options = vec![
```
levkk, line 4: These should be passed in through the context, not hardcoded.

pgml-dashboard/src/utils/url_parser.rs
```diff
@@ -0,0 +1,128 @@
+use lazy_static::lazy_static;
+use regex::Regex;
+
+lazy_static! {
+    static ref DOMAIN_REGEX: Regex =
+        Regex::new(r"^https?://((localhost:8000)|(.*\.?postgresml\.org))").unwrap();
```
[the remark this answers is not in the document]
montanalow: I think we should use the url crate to parse urls, rather than using a regex: https://docs.rs/url/2.4.0/url/struct.Url.html#method.parse

pgml-dashboard/src/utils/url_parser.rs
```diff
@@ -0,0 +1,128 @@
...
+    static ref UUID_REGEX: Regex = Regex::new(".{8}-.{4}-.{4}-.{4}-.{12}").unwrap();
```
montanalow, line 7:
```suggestion
    static ref UUID_REGEX: Regex = Regex::new("[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}").unwrap();
```

pgml-dashboard/src/utils/url_parser.rs
```diff
@@ -0,0 +1,128 @@
...
+    // Determines if there is a uuid in a url.
+    pub fn has_uuid(uri: String) -> bool {
+        match UUID_REGEX.find(&uri) {
+            Some(_) => true,
+            _ => false,
+        }
+    }
+
+    // Replaces the first uuid with a new uuid in a url.
+    pub fn replace_path_uuid(path: String, uuid: String) -> String {
```
montanalow, line 40: I don't see these uuid methods used anywhere.

pgml-dashboard/src/templates/mod.rs
chillenberger: I changed this so the links are in context and set in guards and fairings.

pgml-dashboard/src/utils/url_parser.rs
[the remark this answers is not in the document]
chillenberger: I go rid of the regex and used https://docs.rs/url/2.4.0/url/struct.Url.html#method.parse.  I added some tests to verify we can parse the urls levkk is concerned about.

pgml-dashboard/templates/components/github_icon.html
chillenberger: I am not sure how this got taken out.  I put it back in.  We still need to conditionally show starts since the webapp does not show the stars.

pgml-dashboard/templates/layout/nav/top_web_app.html
chillenberger: I changed this so it is passed in through context. I will need to come back through and move the rest of conditional links to context also.

pgml-dashboard/src/utils/url_parser.rs
chillenberger: This is used in src/api/cluster/mod line 223 of hypercloud repo.  In fact, since a user can only have one local db, we can move the entire utility to the other repo.

pgml-dashboard/src/utils/url_parser.rs
```diff
@@ -1,30 +1,43 @@
 use lazy_static::lazy_static;
 use regex::Regex;
+use url::Url;
```
levkk: I'd be curious to know why we need a url parser. I think Rocket should be able to provide you with everything you need to extract the uuid or anything else you need.

PR #837, opened 2023-07-18
float compare with tolerance for Apple M2

kczimm:
Added floating point comparison using tolerances to failing tests on Apple M2 silicon.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
             ~/.cargo
             pgml-extension/target
             ~/.pgrx
-          key: ${{ runner.os }}-rust-1-${{ hashFiles('pgml-extension/Cargo.lock') }}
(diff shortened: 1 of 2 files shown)


levkk:
Our CI is broken right now, need to bust the tooling cache: https://github.com/postgresml/postgresml/blob/master/.github/workflows/ci.yml#L36

Just replace `-1-` with `-2-` or something like that, and it should work.

levkk:
Oh yeah I should mention, without the cache the tests take about 45-50 min to run :)

levkk:
Up to you! Could probably do all of them, but if they are not broken, no need to fix it either. Just thinking out loud.

kczimm:
I could be convinced either way. At the moment, I'm leaning towards minimal changes to pass tests.

levkk:
Sounds good to me.

PR #838, opened 2023-07-18
fix clippy lints

kczimm:
Fixed clippy warnings. I'm not sure if these changes are of general interest or if they are viewed as nitpicking. Would love the feedback.


pgml-extension/src/orm/snapshot.rs
```diff
@@ -141,10 +141,10 @@ impl Column {
     }
 
     fn nominal_type(pg_type: &str) -> bool {
-        match pg_type {
-            "bpchar" | "text" | "varchar" | "bpchar[]" | "text[]" | "varchar[]" => true,
-            _ => false,
-        }
+        matches!(
```
levkk, line 144: This is cool

pgml-extension/src/orm/snapshot.rs
```diff
@@ -283,16 +279,18 @@ impl Column {
 
         let mut modes = Vec::new();
         statistics.distinct = 0; // necessary reset before array columns clobber
-        for &value in &data {
+        for (i, &value) in data.iter().enumerate() {
```
levkk, line 282: Why would it suggest that? `i` is not used, is it?

levkk:
Cool! Can't argue with clippy, this is definitely better than it was.

pgml-extension/src/orm/snapshot.rs
kczimm: It's used on line 307 I believe when calculating `v`. Previously it was used as an explicit counter outside of the for loop. https://rust-lang.github.io/rust-clippy/master/index.html#/explicit_counter_loop

levkk: That's funny, wonder how we missed that !

montanalow:
At some point we should add clippy back to CI, as it's nice for everyone to have an idiomatic standard.

kczimm:
Should that hypothetically be a separate PR or should I try to add it in here?

montanalow:
I don't think it's critically urgent, or should block this PR.

PR #846, opened 2023-07-20
Mac M1/M2 fixes

levkk:
Tell Rust where libomp and libopenblas are hiding on Mac OS M1+. No need to link anything into `/usr/local/lib` anymore.


kczimm:
This is great. Did you also have to direct CMake through env variables to the clang installed via Homebrew instead of the system install?

levkk:
Yup: https://github.com/postgresml/rust-xgboost/commit/3d4bd10b70117b94367d3c300d233252a204061d

PR #868, opened 2023-07-26
pgml.python_pip_freeze()

levkk:
1. Add `pgml.python_pip_freeze()` to help debug Python package versions used by PostgresML.
2. Add missing dependency installation for macOS.
3. Add test for missing migration file.

### DIFF
--- a/pgml-extension/src/bindings/venv.py
+++ b/pgml-extension/src/bindings/venv.py
@@ -26,13 +26,9 @@ def activate_venv(venv):
 
 
 def freeze():
-    packages = []
     try:
         from pip._internal.operations import freeze
     except ImportError: # pip < 10.0
         from pip.operations import freeze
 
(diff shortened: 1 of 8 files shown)


pgml-extension/src/bindings/venv.py
```diff
@@ -23,3 +23,16 @@ def activate_venv(venv):
...
+    packages = []
+    try:
+        from pip._internal.operations import freeze
+    except ImportError: # pip < 10.0
+        from pip.operations import freeze
+
+    pkgs = freeze.freeze()
+    for pkg in pkgs:
+        packages.append(pkg)
+    return packages
```
kczimm, line 38: Can you just `return list(pkgs)`?

levkk: Oh yeah!

PR #871, opened 2023-07-27
add a bg_worker option to transform

kczimm:


pgml-extension/src/proxy.rs
```diff
@@ -0,0 +1,74 @@
...
+        .set_library(library)
+        .set_restart_time(Some(RESTART_DURATION))
+        .set_argument(socket_addr.into_datum())
+        .load();
+}
+
+/// Send a `Request` to the proxy and wait for a `Reply`.
+pub fn send_request(request: Request) -> io::Result<Reply> {
+    let addr = PROXY_ADDR.get().unwrap();
+    let mut stream = TcpStream::connect(addr)?;
```
levkk, line 65: You want to keep a connection open, otherwise this will be pretty slow (3 roundtrips to establish a connection).

Also I think you want a UNIX socket here.

pgml-extension/src/proxy.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let bytes = request.encode().unwrap();
+    let _ = stream.write(&bytes)?;
+
+    let mut bytes = [0u8; REPLY_SIZE];
+    let _ = stream.read(&mut bytes)?;
```
levkk, line 71: This can be arbitrarily large, since the reply will contain results of a `transform` call.

pgml-extension/src/proxy.rs
```diff
@@ -0,0 +1,74 @@
...
+    match socket_addr.parse::<SocketAddr>() {
+        Ok(addr) => {
+            // UNWRAP: This is the only place it is set and should only be called once
+            PROXY_ADDR.set(addr).unwrap();
+        }
+        Err(e) => error!("could not parse proxy address: {e}"),
+    };
+
+    BackgroundWorkerBuilder::new(NAME)
+        .set_function(function)
```
levkk, line 54: I think this has to  be a no_mangle Rust function: https://github.com/pgcentralfoundation/pgrx/blob/master/pgrx-examples/bgworker/src/lib.rs#L44

kczimm: It does, I tested it. But the no_mangle Rust function will be in a separate cdylib crate specified by `library`, I think.

pgml-extension/src/proxy.rs
kczimm: Yeah I don't know what I was thinking.

pgml-extension/src/proxy.rs
levkk: Cool!

pgml-extension/src/proxy.rs
montanalow: https://github.com/HyperparamAI/zoomies/blob/master/src/lib.rs has a datagram example over udp for unix domain socket.

pgml-extension/src/proxy.rs
levkk: e.g.

```rust
let code = stream.read_u8();
let len = stream.read_i32();
let mut buf = [0u8; len as usize];
stream.read_buffer(&mut buf);

let obj = rmp_serde::from_bytes(&buf).unwrap();
```

levkk: Also would be good to have the socket buffered, so `read_u8()` and others are not actual syscalls.

levkk: You can check if the stream is out of sync by checking that `code` always matches what you'd expect, e.g. `R` for request and `r` for reply.

kczimm: The deserialization of a Reply type would fail if it got out of sync.

pgml-extension/src/proxy.rs
kczimm: If we keep a connection open and reuse it, then we won't get parallel use of the proxy service because we are serializing through the single shared connection. Do we need a connection pool?

montanalow: We don't need a pool in the sense I'm familiar with, but we do need an independent response socket (or some other individually addressed mechanism) for each process that is sending messages on the primary bg_worker socket that also acts as a work queue, so client foreground processes can wait for their individual responses.

i.e. every process sends their request, along with a response addressing mechanism (their process id) to the bg_worker on a shared socket with multiple datagram writers (connection processes) and a single datagram reader (background worker). The bg_worker then connects to a new response socket per process, where only the sender is listening the first time it gets a request from the foreground process.

levkk: The client (pgml) should maintain an open connection. Each new Postgres connection process will open its own connection. The server (bg worker) will listen on the socket and allocate a new connection for each new client.

kczimm: > The client (pgml) should maintain an open connection. Each new Postgres connection process will open its own connection.

I'm not sure I understand this point. But hopefully the new commits will help me understand how far off I am from this thinking.

PR #891, opened 2023-08-03
move postgres logging out of transformer module

kczimm:
- Move postgres logging out of transformer module
  - postgres logging macros silently kill bg_worker process
- remove `device` from kwargs on TypeError in transform and try again

### DIFF
--- a/pgml-extension/src/api.rs
+++ b/pgml-extension/src/api.rs
@@ -1,4 +1,3 @@
-use std::collections::HashMap;
 use std::fmt::Write;
 use std::str::FromStr;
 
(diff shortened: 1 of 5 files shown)


montanalow (approved):
This is great. I'd love to see us start using anyhow results, and context throughout more of the crate, so we can provide cleaner error messages to users.

kczimm:
> This is great. I'd love to see us start using anyhow results, and context throughout more of the crate, so we can provide cleaner error messages to users.

In the future, should we move away from custom error types and just use anyhow everywhere? There is also `thiserror` that make the custom error types easier if we want to go that route.

montanalow:
The `error!` macro works nicely, if we have enough context to give a message to the user, we can cleanly abort and allow them to retry with more information. Too often though, we `unwrap` or `expect` without context, and the resulting rust trace/message is not informative or actionable for the SQL client.

PR #919, opened 2023-08-15
New Rocket slashes not the same

levkk:
New version of Rocket urls that end with `/` and which don't are no longer the same. Nice!


montanalow:
We'll need to validate this change with our google SEO dashboard for any links this breaks, and potentially add manual redirects for any links that have escaped to the wild.

levkk:
We never used URLs that ended with /, just redirected them to the no-trailing slash version. This ensures that we don't provide any links inside the dashboard that has the trailing slash. Inside our web app, we have an automatic redirect setup already.

PR #923, opened 2023-08-15
Speed up releae

levkk:


levkk:
This is going to take a few iterations, Github needs the worklows in master before I can iterate on them in a branch. So, here we go.

PR #925, opened 2023-08-15
add Result to lazy pymodules then trickled up...

### LINKED ISSUE #906: Any crash in pgml.train poisons global
Any error in `pgml.train()` poisons the `Lazy` instance of global state, e.g.

```
select pgml.train('test user', 'classification', 'pgml.digits', 'target');

[...]

ERROR:  Lazy instance has previously been poisoned
```

kczimm:
Attempting to fix #906 when Lazy PyModules fail to initialize by returning a Result. Those Results can trickle all the way up to the API which I did in many cases.

### DIFF
--- a/pgml-extension/src/api.rs
+++ b/pgml-extension/src/api.rs
@@ -13,16 +13,25 @@ use serde_json::json;
 use crate::bindings::sklearn::package_version;
 use crate::orm::*;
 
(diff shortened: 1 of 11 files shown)


levkk (approved):
Very cool! I think this will work!

PR #982, opened 2023-09-05
Add tabs & table

levkk:
### Table
image

### Tabs
image


pgml-dashboard/build.rs
```diff
@@ -3,7 +3,6 @@ use std::process::Command;
 
 fn main() {
     println!("cargo:rerun-if-changed=migrations");
-    println!("cargo:rerun-if-changed=src");
```
levkk, line 6: @chillenberger This rebuilt the app every time we changed css which was kinda slow. Instead, I'm running:

```
cargo watch  --exec 'pgml-components bundle'  --watch src/  --watch static/  --ignore bundle.*.*
```

in a separate terminal.

PR #1013, opened 2023-09-13
update huggingface dependencies and calls

montanalow:


levkk:
This reminds me, when we release the new version into APT, we need to release a new version of this package: https://github.com/postgresml/postgresml/actions/workflows/ubuntu-postgresml-python-package.yaml

I split them up to help with upgrades. Most of the time, we're only changing code, and the Python dependencies stay the same; upgrading just the .so is much faster than downloading all the Python dependencies all over again.

PR #1029, opened 2023-09-25
fix pgml.tune

### LINKED ISSUE #1028: pgml.tune throws an error
I'm having problems with pgml.tune as well.

I tried the sample as is, but I think its outdated:
```
SELECT pgml.tune(
    'IMDB Review Sentiment',
    task => 'text-classification',
    relation_name => 'pgml.imdb',
    y_column_name => 'label',
    model_name => 'distilbert-base-uncased',
    hyperparams => '{
        "learning_rate": 2e-5,
        "per_device_train_batch_size": 16,
        "per_device_eval_batch_size": 16,
        "num_train_epochs": 1,
        "weight_decay": 0.01
    }',
    test_size => 0.5,
    test_sampling => 'last'
);
```
results in: 
```
ERROR:  only text type columns are supported
```

So I created a view to make them both text:
```
CREATE OR REPLACE VIEW imdb_text AS
SELECT text AS "text", cast(label as text) AS "label"
FROM pgml.imdb
LIMIT 1000;
```
resulted in: 
```
UnboundLocalError: local variable 'max_length' referenced before assignment
```

kczimm:
Attempts to fix #1028

### DIFF
--- a/pgml-extension/src/bindings/lightgbm.rs
+++ b/pgml-extension/src/bindings/lightgbm.rs
@@ -24,18 +24,21 @@ impl std::fmt::Debug for Estimator {
     }
 }
 
-pub fn fit_regression(dataset: &Dataset, hyperparams: &Hyperparams) -> Result<Box<dyn Bindings>> {
+pub fn fit_regression(
+    dataset: &Dataset<f32>,
+    hyperparams: &Hyperparams,
+) -> Result<Box<dyn Bindings>> {
     fit(dataset, hyperparams, Task::regression)
 }
 
 pub fn fit_classification(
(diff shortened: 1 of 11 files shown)


kczimm:
Assuming the column types are appropriate due to something like

```sql
CREATE OR REPLACE VIEW imdb_text AS
SELECT text AS "text", cast(label as text) AS "label"
FROM pgml.imdb
LIMIT 1000;
```

it now silently aborts. From the log:

```
LOG:  server process (PID 55296) was terminated by signal 6: Abort trap: 6
```

montanalow:
Why does it abort?

PR #1052, opened 2023-10-04
gitbook assets

montanalow:
TODO
- configuration to deploy docs


pgml-dashboard/src/api/docs.rs
```diff
@@ -24,54 +24,36 @@ async fn search(query: &str, index: &State<markdown::SearchIndex>) -> ResponseOk
...
-            NavLink::new("Snapshots").href("/docs/guides/schema/snapshots"),
-            NavLink::new("Projects").href("/docs/guides/schema/projects"),
-            NavLink::new("Deployments").href("/docs/guides/schema/deployments"),
-        ]),
-    ];
-
-    render(cluster, &path, guides, "Guides", &Path::new("docs")).await
+use rocket::fs::NamedFile;
+
+#[get("/docs/guides/.gitbook/assets/<path>", rank = 10)]
```
levkk, line 29: You can use https://api.rocket.rs/v0.5-rc/rocket/fs/struct.FileServer.html

pgml-dashboard/src/components/navbar_web_app/template.html
```diff
@@ -48,7 +48,7 @@
             </li>
   
             <li class="nav-item d-flex align-items-center">
-              <a class="nav-link p-lg-0" href="/docs/guides/setup/quick_start_with_docker">Docs</a>
+              <a class="nav-link p-lg-0" href="/docs/guides/README">Docs</a>
```
levkk, line 51: Not very SEO friendly imo

pgml-dashboard/src/utils/markdown.rs
```diff
@@ -1,45 +1,44 @@
...
     }
 }
 
 impl HeadingAdapter for MarkdownHeadings {
     fn enter(&self, meta: &HeadingMeta) -> String {
-        let id = self
-            .header_counter
-            .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
-        let id = format!("header-{}", id);
+        let id = meta.content.to_case(convert_case::Case::Kebab);
```
levkk, line 41: Not unique enough, we have duplicate headers in blogs and docs.

pgml-dashboard/src/api/docs.rs
montanalow: Can't use the fileserver because we need the path to not be exact, but rather the relative gitbook path.

pgml-dashboard/src/components/navbar_web_app/template.html
montanalow: Yeah, but we can't control this lever in gitbook... and then again... it is kinda devvy.

pgml-dashboard/src/utils/markdown.rs
montanalow: We should dedup them. URLS are part of the UX, and we're also losing control of this to gitbook.

levkk:
Do you think we should place these into `/docs2` and hide them from Google until we finish our work on them? To me the current docs feel more usable than these ones.

PR #1054, opened 2023-10-05
Chatbot page is almost ready to go

SilasMarvin:


pgml-dashboard/src/components/star/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+    );
+    map.insert(
+        "party",
+        include_str!("../../../static/images/icons/stars/party.svg"),
+    );
+    map
+});
+
+impl Star {
+    pub fn new<S: ToString, I: Into<Option<S>>>(color: &str, content: S, id: I) -> Star {
```
levkk, line 29: I discovered this the other day:

```
fn something(content: impl ToString)
```

Seems cleaner than a generic notation.

pgml-dashboard/src/components/star/star_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from '@hotwired/stimulus'
```
levkk: You can delete empty controllers, bundle just won't include them.

pgml-dashboard/static/js/utilities/toast.js
```diff
@@ -1,35 +1,35 @@
 function createToast(message) {
-    const toastElement = document.createElement('div');
-    toastElement.classList.add('toast', 'hide');
-    toastElement.setAttribute('role', 'alert');
-    toastElement.setAttribute('aria-live', 'assertive');
-    toastElement.setAttribute('aria-atomic', 'true');
+  const toastElement = document.createElement("div");
```
levkk, line 2: Do you have a linter we could all install? It would be nice to have one tbh...

pgml-dashboard/src/components/star/star_controller.js
SilasMarvin: Yes I just removed this!

pgml-dashboard/Cargo.toml
```diff
@@ -45,3 +45,5 @@ pgvector = { version = "0.2.2", features = [ "sqlx", "postgres" ] }
 console-subscriber = "*"
 glob = "*"
 pgml-components = { path = "../packages/pgml-components" }
+reqwest = { version = "0.11.20", features = ["json"] }
+pgml = { version = "0.9.2", path = "../pgml-sdks/pgml/" }
```
levkk, line 49: Love monorepos! Things are just easier.

pgml-dashboard/static/js/utilities/toast.js
SilasMarvin: I use prettier for javascript. I can setup a prettier configuration for our repos that will avoid things like this.

levkk: Please yeah, any one of them works, as long as it's easy to use.

PR #1060, opened 2023-10-10
Added a custom Option type and extended the select and added a few new classes for typography

SilasMarvin:


pgml-dashboard/static/css/scss/base/_typography.scss
```diff
@@ -11,10 +13,60 @@
...
-.text-black {color: #{$gray-900} !important;}
-.text-white {color: #{$gray-100} !important;}
-.text-soft-white { color: #{$gray-200} !important; }
+.text-black {
+    color: #{$gray-900} !important;
+}
+.text-white {
+    color: #{$gray-100} !important;
+}
+.text-gradient-blue {
```
chillenberger, line 35: Using a mixin for this would prevent the repetition.

chillenberger (approved):

pgml-dashboard/static/css/scss/base/_typography.scss
SilasMarvin: Let me update this. I haven't used mixins before

levkk: classes are better than mixins, mixings generate a ton of repetitive css code, while classes inherit and cascade.

chillenberger: @extend is also an option that minimizes css code (same as having two classes) and reduces the number of classes an element needs to take.  If we do go with two classes here we should have the _typographic class .text-gradient then a utility class for all the gradients so they can be applies as text color, background color, border color ect.

PR #1063, opened 2023-10-11
vllm support

kczimm:
Enables support for vLLM. To use, you must specify the `model` field in the `task` parameter of the `pgml.transform` function **and** you must add `"backend": "vllm"` in the `task` parameters. For example,

```sql
SELECT * FROM pgml.transform(
    task => '{"model":"tiiuae/falcon-7b","backend":"vllm"}'::JSONB,
    inputs => Array['hello']
);
```

A list of supported models for vLLM can be found [here](https://vllm.readthedocs.io/en/latest/models/supported_models.html).

Only one vLLM model can be loaded per client connection process due to a [limitation in vLLM](https://github.com/vllm-project/vllm/issues/565). The first call to `pgml.transform` with a given model will load the model ("cold start"), but subsequent calls will use the cached model. If you change the specified model in the same client connection, the cached model will be replaced with the new one.


levkk:
Rebase on master to get https://github.com/postgresml/postgresml/pull/1102 which should fix the tests.

PR #1077, opened 2023-10-16
Added an accordian and another star

SilasMarvin:


pgml-dashboard/src/components/accordian/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+use pgml_components::component;
+
+// This component will probably not work very well if two are on the same page at once. We can get
+// around it if we include some randomness with the data values in the template.html but that
+// doesn't feel very clean so I will leave this problem until we have need to fix it or a better
+// idea of how to
+#[derive(TemplateOnce, Default)]
+#[template(path = "accordian/template.html")]
+pub struct Accordian {
+    html_contents: Vec<String>,
```
levkk, line 11: Use `pgml_components::Component` so the caller doesn't have to `render_once()` the components.

PR #1094, opened 2023-10-19
add options capability to range input, add typography to site, fix sw…

chillenberger:
- Typography for site
 - Add options to range input
 - Center text in switch input
Screenshot 2023-10-19 at 11 09 39 AM

### DIFF
--- a/pgml-dashboard/static/css/scss/base/_typography.scss
+++ b/pgml-dashboard/static/css/scss/base/_typography.scss
@@ -1,19 +1,62 @@
 // all other displays are default bootstrap styling
 .display-2 {
-    font-weight: 700;
+    font-weight: $font-weight-bold;
     font-size: 4rem;
     line-height: 80px;
 }
 
-// TODO: Go through html and ensure headers use appropriate header class, header and class do not need to match.
-// .h1 {font-weight: 700; font-size: 64px; line-height: 72px;}
-// .h2 {font-weight: 700; font-size: 48px; line-height: 54px;}
-// .h3 {font-weight: 700; font-size: 40px; line-height: 46px;}
-// .h4 {font-weight: 700; font-size: 32px; line-height: 40px;}
-// .h5 {font-weight: 700; font-size: 28px; line-height: 34px;}
-// .h6 {font-weight: 700; font-size: 24px; line-height: 30px;}
+.h1-big {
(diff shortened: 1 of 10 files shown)


SilasMarvin (approved):

pgml-dashboard/static/css/scss/base/_typography.scss
```diff
@@ -1,19 +1,62 @@
...
-// TODO: Go through html and ensure headers use appropriate header class, header and class do not need to match.
-// .h1 {font-weight: 700; font-size: 64px; line-height: 72px;}
-// .h2 {font-weight: 700; font-size: 48px; line-height: 54px;}
-// .h3 {font-weight: 700; font-size: 40px; line-height: 46px;}
-// .h4 {font-weight: 700; font-size: 32px; line-height: 40px;}
-// .h5 {font-weight: 700; font-size: 28px; line-height: 34px;}
-// .h6 {font-weight: 700; font-size: 24px; line-height: 30px;}
+.h1-big {
+    font-weight: $font-weight-bold; font-size: 80px; line-height: 84px;
+    @include media-breakpoint-down(md) {
```
levkk, line 10: Is this right? I remember last time we did this, the fonts were too small on mobile, not too big.

levkk: I believe you're also looking for "display" classes which are the headers but "big": https://getbootstrap.com/docs/5.3/content/typography/#display-headings

levkk: On another note, I don't think we should be using pixels anymore, but rems.

chillenberger: No before they were too big on mobile.  I'll send you the slack conversation.

chillenberger: I agree about this needing to be in rem and use display.  Honestly this can be redone the bootstap way (setitng the scss variables).  I just wanted a quick fix to make things look right. I can put it in rem the next change I touch dashboard.

levkk: We should adjust our typography accordingly I think. Pixels are not responsive if I remember correctly, that's why the industry switched to ems and then to rems.

PR #1102, opened 2023-10-19
Update to pgrx 0.11.0 & Postgres 16

levkk:
Reverts postgresml/postgresml#1101


maparent:
Thank you! Could I ask for one more favour? Projects that want to use this would benefit from a version bump.

montanalow:
> Thank you! Could I ask for one more favour? Projects that want to use this would benefit from a version bump.

2.7.12 is out now.

PR #1107, opened 2023-10-21
2.7.12

montanalow:


packages/postgresml-python/build.sh
```diff
@@ -7,7 +7,7 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 deb_dir="/tmp/postgresml-python/deb-build"
 major=${1:-"14"}
 
-export PACKAGE_VERSION=${1:-"2.7.10"}
+export PACKAGE_VERSION=${1:-"2.7.12"}
```
levkk, line 10: You don't need to release a new version of this package unless you've changed python deps since the last version.

montanalow: It seems like a nice convenience for end users, if all package versions for the project align.

levkk: That's not how this works or why we split the python deps into their own package. This allows for us to upgrade the extension much quicker without having to download and install gigabytes of dependencies.

montanalow: That split is causing issues with consistent updates.

levkk: #1113

levkk: As I wrote the docs I remembered a couple more reasons why I split those packages and I'm pretty confident in my decision now. On the `postgresml/postgresml` side of things, we're pretty covered I think, as long as we follow release procedures (https://github.com/postgresml/postgresml/pull/1013#issuecomment-1721879852)

montanalow: Having code that implements the procedure seems like the next step after a manual procedure is documented.

PR #1111, opened 2023-10-21
Interactable ToC

montanalow:
Various fixes for the interactivity of the ToC

- Use human readable anchors that refer to the content for stable external links and seo
- Add a visual indicator that headings are clickable (#)
- Update the window.location when the ToC is clicked for increased discoverability
- Fix scrollspy being re-appended every time it activates... recursively adding itself until it's firing exponentially
- Fix deep links to headers being obscured by our sticky nav

### DIFF
--- a/pgml-dashboard/src/templates/docs.rs
+++ b/pgml-dashboard/src/templates/docs.rs
@@ -1,5 +1,6 @@
 //! Documentation and blog templates.
 use sailfish::TemplateOnce;
+use convert_case::Casing;
 
 use crate::utils::markdown::SearchResult;
 
@@ -78,12 +79,10 @@ impl TocLink {
     ///
     /// * `title` - The title of the link.
     ///
-    pub fn new(title: &str, counter: usize) -> TocLink {
-        let id = format!("header-{}", counter);
-
+    pub fn new(title: &str) -> TocLink {
         TocLink {
             title: title.to_string(),
-            id,
+            id: title.to_case(convert_case::Case::Kebab),
             level: 0,
         }
     }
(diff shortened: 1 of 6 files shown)


pgml-dashboard/src/templates/docs.rs
```diff
@@ -78,12 +79,10 @@ impl TocLink {
...
     /// * `title` - The title of the link.
     ///
-    pub fn new(title: &str, counter: usize) -> TocLink {
-        let id = format!("header-{}", counter);
-
+    pub fn new(title: &str) -> TocLink {
         TocLink {
             title: title.to_string(),
-            id,
+            id: title.to_case(convert_case::Case::Kebab),
```
levkk, line 85: Could you add a unique id here, maybe:

```rust
id: format!("{}-{}", title, counter),
```

This accomplishes both your goal of making the IDs user-friendly and also ensuring that duplicated headers (which are very common in our docs and blog) don't break the ToC.

pgml-dashboard/static/js/docs-toc.js
```diff
@@ -4,15 +4,23 @@ import {
 
 export default class extends Controller {
   connect() {
-    this.scrollSpyAppend();
-  }
-
-  scrollSpyAppend() {
     const spy = new bootstrap.ScrollSpy(document.body, {
```
levkk, line 7: I'm surprised this doesn't get garbage collected immediately. Perhaps the plugin attaches itself to the element and exists until that element is removed.

pgml-dashboard/src/templates/docs.rs
levkk: At the very least if you don't like having document-global unique IDs, maybe have a `HashSet` + counter that checks if the header already exists and if it does, append a counter to it? This will replicate the implementation mkdocs has, where if the headers are unique, they are using kebab case, and if not, it starts counting, e.g. `#non-unique-header`, `#non-unique-header-1`, `#non-unique-header-2`, etc.

PR #1113, opened 2023-10-22
Document packages and release process

levkk:
Document packages we use to distribute PostgresML. Document release process followed for new packages.


levkk:
@montanalow Could you take a look and let me know if I'm missing some details. This should serve as both reference for us and onboarding documentation for new contributors.

PR #1114, opened 2023-10-23
Mutable SQL commands should be used with Spi::update().

higuoxing:
Commands like 'INSERT', 'CREATE TABLE' are modifying database objects. They should be used with 'Spi::update()'[^1][^2].

Besides, one of the unnecessary calling of '.clone()' is removed.

[^1]: https://www.postgresql.org/docs/current/spi-spi-execute.html
[^2]: https://github.com/pgcentralfoundation/pgrx/blob/64fbc2e2efb06212046082ffcde2e794ccb4d48a/pgrx/src/spi/client.rs#L60

Co-Authored-By: Chen Mulong <chenmulong@gmail.com>


montanalow:
Thanks for this fix!

PR #1125, opened 2023-10-26
Mistral AI and GGUF support

levkk:
- Update deps for Mistral
- Add support for GGUF (wasn't hard)

### DIFF
--- a/pgml-extension/src/bindings/transformers/transformers.py
+++ b/pgml-extension/src/bindings/transformers/transformers.py
@@ -197,7 +197,7 @@ def create_pipeline(task):
         lower = model_name.lower()
     else:
         lower = None
-    if lower and "-ggml" in lower:
+    if lower and ("-ggml" in lower or "-gguf" in lower):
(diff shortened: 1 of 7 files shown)


kczimm (approved):

pgml-extension/src/bindings/transformers/transformers.py
```diff
@@ -197,7 +197,7 @@ def create_pipeline(task):
         lower = model_name.lower()
     else:
         lower = None
-    if lower and "-ggml" in lower:
+    if lower and ("-ggml" in lower or "-gguf" in lower):
```
levkk, line 200: @kczimm these run on the CPU, so could be useful for local dev on a M1. I'll try on my end just to be sure (on my desktop right now)

SilasMarvin (approved):

pgml-extension/src/bindings/transformers/transformers.py
levkk: Worked on my M1 after uninstalling `deepspeed`.

PR #1143, opened 2023-11-02
Invalid forms style

levkk:


pgml-dashboard/static/css/scss/components/_forms.scss
```diff
@@ -257,6 +257,19 @@
...
+    &.is-invalid {
+        &:focus {
+            box-shadow: none;
+            border-width: 2px;
+        }
+    }
+}
+
+.invalid-feedback {
+    --bs-danger-text: #{$peach-shade-100};
```
chillenberger, line 270: As of now all the error colors should be the same and use the variable $error set to $peach-shade-100.  So, it might be best to use $error and deviate later if needed.  Same comment for the above $form-feedback-invalid-color:

PR #1145, opened 2023-11-02
Add streaming

SilasMarvin:


pgml-extension/src/bindings/transformers/mod.rs
```diff
@@ -52,6 +52,43 @@ pub fn transform(
     Ok(serde_json::from_str(&results)?)
 }
 
+pub fn transform_stream(
```
montanalow, line 55: Why not replace the current transform implementation with the streaming version?

pgml-extension/src/bindings/transformers/mod.rs
```diff
@@ -52,6 +52,43 @@ pub fn transform(
...
+) -> Result<Py<PyAny>> {
+    crate::bindings::python::activate()?;
+
+    whitelist::verify_task(task)?;
+
+    let task = serde_json::to_string(task)?;
+    let args = serde_json::to_string(args)?;
+    let inputs = serde_json::to_string(&inputs)?;
+
+    Python::with_gil(|py| -> Result<Py<PyAny>> {
```
montanalow, line 68: Can probably move all this into the transformers module just for organization.

pgml-extension/src/bindings/transformers/transformers.py
```diff
@@ -110,6 +112,31 @@ def __call__(self, inputs, **kwargs):
...
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        if len(self.done_data) > 0:
+            return self.done_data.pop(0)
+        elif self.done:
+            raise StopIteration
+        time.sleep(0.1)
```
montanalow, line 136: Why does this need to sleep?

montanalow (commented):
Very exciting!

pgml-extension/src/bindings/transformers/transformers.py
SilasMarvin: If I don't have it sleep I get:
```ERROR:  RecursionError: maximum recursion depth exceeded```

pgml-extension/src/bindings/transformers/mod.rs
SilasMarvin: I guess if the user calls it without a limit or cursor, they do the same exact thing. I'm not sure what the overhead of dipping into python for every token is. If that isn't an issue I totally can!

pgml-extension/src/bindings/transformers/transformers.py
levkk: That's weird...

pgml-extension/src/bindings/transformers/mod.rs
montanalow: I love the focus on performance. I think we can keep the two different implementations under a single API function, with a fork in the logic depending on the params. i.e. the caller shouldn't have to think about that as a high level split, and we shouldn't have to duplicate all the documentation for the function, just the change in behavior based on those arguments.

pgml-extension/src/bindings/transformers/transformers.py
montanalow: This is because we hit the recursion depth before a new token is added to done_data. I'd suggest moving to a while loop rather than recursing since Python is a ... does not support tail recursion.

montanalow: After moving to a while loop, I think it's worth a big fat comment warning that we are spin locking in this thread, and sleeping for some amount of time on each iteration in a pure `else` case, and sleeping for some amount of time may be a better cost/latency tradeoff.

montanalow: Is it possible to lock/wait/notify with postgres in this case?

levkk: https://docs.python.org/3.10/library/queue.html

https://stackoverflow.com/questions/14508906/sending-messages-between-class-threads-python

PR #1146, opened 2023-11-06
Support custom huggingface and sentence-transformers cache

liuxueyang:
Currently, sentence-transformers use different cache path from huggingface by default. This makes it difficult for users to copy cached models if they want to run pgml in offline mode.

This PR adds two GUCs `pgml.cache` and `pgml.offline`. If user sets `pgml.cache`, then the `HF_HOME` will be set to `pgml.cache/huggingface` and `SENTENCE_TRANSFORMERS_HOME` will be set to `pgml.cache/torch`.

The sentence-transformers still use the old version of cache file layout: https://github.com/UKPLab/sentence-transformers/pull/2339


pgml-extension/src/bindings/transformers/mod.rs
```diff
@@ -1,15 +1,22 @@
...
 use anyhow::{anyhow, bail, Context, Result};
+use lazy_static::*;
 use pgrx::*;
+use pyo3::ffi;
 use pyo3::prelude::*;
 use pyo3::types::PyTuple;
 use serde_json::Value;
+use std::cmp::Ordering;
+use std::env;
+use std::sync::RwLock;
```
levkk, line 15: Use `parking_lot`, it doesn't poison the lock on panic.

pgml-extension/src/bindings/transformers/mod.rs
```diff
@@ -18,11 +25,18 @@ pub mod whitelist;
 
 create_pymodule!("/src/bindings/transformers/transformers.py");
 
+lazy_static! {
+    // Record the previous applied ENVs.
+    static ref ENVS_APPLIED: RwLock<Box<BTreeMap<&'static str, String>>> =
```
levkk, line 30: I'm struggling to understand the purpose of this. Why do we need to record what environment variables we've applied? Can't we just set them every time or they can only be set on initialization before HF has been imported?

pgml-extension/src/bindings/transformers/mod.rs
```diff
@@ -389,3 +408,88 @@ pub fn clear_gpu_cache(memory_usage: Option<f32>) -> Result<bool> {
         Ok(success)
     })
 }
+
+// Called before hugginface python APIs. Setup ENVs for HuggingFace. See
+// https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hfhuboffline
+pub fn set_env() {
+    let envs_to_apply = guc::gen_hf_env_map();
+    let py_inited = unsafe { ffi::Py_IsInitialized() != 0 };
```
levkk, line 416: This works with PyO3? We've made sure that PyO3 is loaded into local area in our .so: https://github.com/postgresml/postgresml/blob/master/pgml-extension/ld.map#L7

pgml-extension/src/bindings/transformers/mod.rs
```diff
@@ -389,3 +408,88 @@ pub fn clear_gpu_cache(memory_usage: Option<f32>) -> Result<bool> {
...
+
+    {
+        // This block can not be removed. It's used to drop read lock
+        // read lock held
+        let envs_current = ENVS_APPLIED.read().unwrap();
+
+        if py_inited {
+            if envs_to_apply.cmp(&envs_current) != Ordering::Equal {
+                // Python had been initialized and GUCs changed. Report warning and do nothing.
```
levkk, line 425: Why?

pgml-extension/src/bindings/transformers/mod.rs
```diff
@@ -69,6 +83,7 @@ pub fn embed(
     inputs: Vec<&str>,
     kwargs: &serde_json::Value,
 ) -> Result<Vec<Vec<f32>>> {
+    set_env();
```
montanalow, line 86: What's the latency/lock contention cost of accessing the GUC on every call of embed?

pgml-extension/src/bindings/transformers/mod.rs
liuxueyang: The purpose of the recorded map is to warn users if the GUCs are changed and the changes can not take effect in the current session. This prevents users from downloading/caching the models in the old cache path.

levkk: Why can't we set the env using the latest GUC value every time?

liuxueyang: The value of the latest GUC can not be seen by the old psql session started before changing the GUC.

pgml-extension/src/bindings/transformers/mod.rs
liuxueyang: Fixed

pgml-extension/src/bindings/transformers/mod.rs
liuxueyang: The cost is acquiring the mutex lock, building a treemap in `gen_hf_env_map`, and comparing two treemaps. It seems a little inefficient.

liuxueyang:
The advantage of adding these GUCs for environment variables is that users can change the huggingface home and sentence-transformer cache path without restarting the Postgres.

montanalow:
Having to check 2 GUCs on every invocation seems potentially heavy handed without understanding the motivation for why these settings would need to be more dynamic than a server restart.

liuxueyang:
If we don't need that, maybe we can simply initialize these two environments in PG_init (`orm::project::init`)

montanalow:
> If we don't need that, maybe we can simply initialize these two environments in PG_init (`orm::project::init`)

This sounds like a good compromise to me. Is PG_init called on every connection process fork? Would the behavior then be that new connections get the updated conf values, or is this something that can only be changed at server start?

montanalow:
It'd be nice to validate our approach here also resolves #1147.

higuoxing:
> This sounds like a good compromise to me. Is PG_init called on every connection process fork? Would the behavior then be that new connections get the updated conf values, or is this something that can only be changed at server start?

+1 to the idea that initialize GUC values in `_PG_init`.

> Is PG_init called on every connection process fork?

`_PG_init`  is called each time the shared library gets loaded. Since the `pgml` is added to the `shared_preload_libraries` list, `PG_init` is called on every connection process fork.

> Would the behavior then be that new connections get the updated conf values

Normally, the server processes forked for new connections use the default value (`boot_val`).

```
extern void DefineCustomStringVariable(const char *name,
					   const char *short_desc,
					   const char *long_desc,
					   char **valueAddr,
					   const char *bootValue,
					   GucContext context,
					   int flags,
					   GucStringCheckHook check_hook,
					   GucStringAssignHook assign_hook,
					   GucShowHook show_hook) pg_attribute_nonnull(1, 4);
```

> is this something that can only be changed at server start

There're several flags we can use in `DefineCustomXXXVariable` to allow GUC values to be changed only at the time of server starting[^1].

[^1]: https://github.com/postgres/postgres/blob/8f4a1ab471e6a258881e3d7f5883705714cde7e1/src/include/utils/guc.h#L53C1-L53C1

liuxueyang:
I've refactored the code to initialize the environment variables in `_PG_init` only once.

liuxueyang:
If we want to benefit from the GUC flags we can define the GUCs using functions in GucRegistry, for example:

```
GucRegistry::define_string_guc(
        "pgml.cache",
        "pgml cache directory",
        "This directory will serve as huggingface home and pytorch cache",
        &GP_GUCS.pgml_cache,
        GucContext::Suset,
        GucFlags::default(),
    );
```

If we need this, we can add it to this PR

liuxueyang:
@montanalow Since pgml is in `shared_preload_libraries`, so pgml is loaded exactly once on postmaster start. `_PG_init` is called when the library is loaded. So these environment variables will only be changed on server start.

montanalow (approved):
This looks good on my side. @levkk will we need any additional configuration in our cloud after this is merged?

pgml-extension/src/config.rs
```diff
@@ -1,9 +1,36 @@
...
+static ENV_SENTENCE_TRANSFORMERS_HOME: &str = "SENTENCE_TRANSFORMERS_HOME";
+static ENV_HF_HUB_OFFLINE: &str = "HF_HUB_OFFLINE";
+
+struct Guc {
+    pgml_cache: GucSetting<Option<&'static CStr>>,
+    pgml_offline: GucSetting<bool>,
+}
+
+lazy_static! {
+    static ref GUC: Guc = Guc {
```
montanalow, line 28: Since this is now only used once at server start, this can be a transient variable to set the env, right? We don't need to keep it around in a lazy_static?

liuxueyang: The reason that we use functions in GucRegistry is that we can define bool GUC and add descriptions for GUC variables.

liuxueyang:
Update: The sentence-transformers update cache format to git style via hf_hub_download now: https://github.com/UKPLab/sentence-transformers/pull/2345

btw, is there any progress on this PR?

PR #1149, opened 2023-11-08
add reindex to syntax highlight

chillenberger:
- I believe we need reindex in the syntax highlight 
befoer : 
Screenshot 2023-11-07 at 7 17 34 PM

after: 
Screenshot 2023-11-07 at 7 17 43 PM


montanalow:
I think we should move toward our client side syntax highlighting for all code to have a single consistent presentation.

PR #1164, opened 2023-11-16
Added support for llama and mistral GPTQ models

SilasMarvin:
It is required that Mistral models use Transformers, and Llama models can use Transformers. I'm guessing Transformers is more stable so I also have Llama use Transformers and not AutoGPTQ.

### DIFF
--- a/pgml-extension/src/bindings/transformers/transformers.py
+++ b/pgml-extension/src/bindings/transformers/transformers.py
@@ -294,13 +294,16 @@ def create_pipeline(task):
     ensure_device(task)
     convert_dtype(task)
     model_name = task.get("model", None)
+    model_type = None
+    if "model_type" in task:
+        model_type = task["model_type"]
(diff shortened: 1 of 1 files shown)


levkk (approved):
Cool

montanalow:
Did gptq support get merged into transformers? IIRC it wasn't originally supported. Can/should we remove auto-gptq special casing alltogether?

SilasMarvin:
> Did gptq support get merged into transformers? IIRC it wasn't originally supported. Can/should we remove auto-gptq special casing alltogether?

I'm not sure. I've only tested with these two models, but I'll look into it!

SilasMarvin:
> Did gptq support get merged into transformers? IIRC it wasn't originally supported. Can/should we remove auto-gptq special casing alltogether?

Yes great call.

https://github.com/postgresml/postgresml/pull/1165

Issue #1161, opened 2023-11-16
PostgresML is spawning too many threads.


higuoxing:
Recently, we found that PostgresML was spawning too many threads and threads is waiting for each other. No deadlock but low performance.

## Root cause

The root cause is that PyTorch spawns too many threads, which cause severe CPU contention.

This can be solved by limiting the degree of parallelism with e.g.,

```
import torch
torch.set_num_threads(4)
```

The root cause and workaround is found and provided by @xuebinsu

Probably we can add some GUC parameter to allow user tunning the performance?

## Reproducing steps

Run test without limiting the number of threads. (Finished in 273.66s)

```
$ unset OMP_NUM_THREADS
$ cargo pgrx test pg16
...
test result: ok. 57 passed; 0 failed; 25 ignored; 0 measured; 0 filtered out; finished in 273.66s
```

Run test by limiting the number of threads that openmp can spawn with `OMP_NUM_THREADS=1`. (Finished in 7.81s)

```
$ export OMP_NUM_THREADS=1
$ cargo pgrx test pg16
...
test result: ok. 57 passed; 0 failed; 25 ignored; 0 measured; 0 filtered out; finished in 7.81s
```

## Stack trace

<details>

<summary> Stack trace </summary>

```
process: /proc/180629/mem
thread: 0x7ff82594f740, lwp: 180629, type: 1
#0  0x00007ff82518ca1e in gomp_team_barrier_wait_end()+318 in /usr/lib/libgomp.so.1 at futex.h:111
#1  0x00007ff82518b35d in gomp_team_end()+44 in /usr/lib/libgomp.so.1 at team.c:956
#2  0x00007ff719a7935f in _ZN7xgboost6common13ParallelFor2dIZNS_4tree16HistogramBuilder13SyncHistogramEPKNS_7RegTreeERKSt6vectorIiSaIiEESB_EUlmNS0_7Range1dEE2_EEvRKNS0_14BlockedSpace2dEiOT_!()+302 in /home/v/.local/lib/python3.11/site-packages/xgboost/lib/libxgboost.so
#3  0x00007ff719a857bd in _ZN7xgboost4tree16HistogramBuilder13SyncHistogramEPKNS_7RegTreeERKSt6vectorIiSaIiEES9_!()+1404 in /home/v/.local/lib/python3.11/site-packages/xgboost/lib/libxgboost.so
[...]

PR #1186, opened 2023-11-25
Added embed2 which returns a table structure

ns1000:
The proposed change added new wrapper for pgml.embed, this will instead of return a single row, return a table structure which is very useful for batch processing strings.

Example usage is
```sql
select * from 
    pgml.embed2('all-MiniLM-L6-v2', (select array_agg(phrase) from (select * from phrases limit 10)));
```
To import the function is as follows:
```sql
CREATE OR REPLACE FUNCTION pgml."embed2"(
    transformer TEXT,
    inputs TEXT[],
    kwargs JSONB DEFAULT '{}'
) RETURNS TABLE (text TEXT, embedding real[]) 
    LANGUAGE c IMMUTABLE STRICT PARALLEL SAFE AS 'MODULE_PATHNAME', 'embed_batch2_wrapper';
```

### DIFF
--- a/pgml-extension/src/api.rs
+++ b/pgml-extension/src/api.rs
@@ -558,6 +558,26 @@ pub fn embed_batch(
     }
 }
 
+#[cfg(all(feature = "python", not(feature = "use_as_lib")))]
+#[pg_extern(immutable, parallel_safe, name = "embed2")]
+pub fn embed_batch2<'a>(
+    transformer: &str,
+    inputs: Vec<&'a str>,
+    kwargs: default!(JsonB, "'{}'"),
+) -> TableIterator<'a, (name!(text, String), name!(embedding, Vec<f32>))> {
+    let rows = match crate::bindings::transformers::embed(transformer, inputs.clone(), &kwargs.0) {
+        Ok(rows) => rows,
+        Err(e) => {
+            error!("{e}");
+        }
+    };
+    TableIterator::new(
+        inputs.into_iter().zip(rows.into_iter()).map(|(text, embedding)| {
+            (text.to_string(), embedding)
+        }),
+    )
+}
+
 /// Clears the GPU cache.
 ///
 /// # Arguments
--- a/pgml-extension/requirements.txt
+++ b/pgml-extension/requirements.txt
@@ -29,3 +29,4 @@ pynvml==11.5.0
 transformers-stream-generator==0.0.4
 optimum==1.13.2
 peft==0.6.2
+pyarrow==11.0.0


pgml-extension/src/api.rs
```diff
@@ -558,6 +558,26 @@ pub fn embed_batch(
     }
 }
 
+#[cfg(all(feature = "python", not(feature = "use_as_lib")))]
+#[pg_extern(immutable, parallel_safe, name = "embed2")]
+pub fn embed_batch2<'a>(
```
montanalow, line 563: My rough thoughts, without running the code on some examples.

I think we should name this `embed_batch_3` with the goal of establishing this as the 3.0 `embed` API, as well as a pattern for releasing 3.0 APIs early as we developing them in an alpha state (with potentially breaking changes, where we completely drop them in 3.1 in favor of the newly established default behavior). 

Your example convinces me that batch APIs should return a table, but I think that table's rows should be `JSONB` with {id, embedding} keys (at least), unless there is a significant performance implication on that front. My thinking is that embedding models are getting more complicated and now some take JSON rather than TEXT for `inputs` including a `prompt`.  It would be nice to have an optional `id` in the input JSON, and if it's not present, then just return the entire input JSON as the `id`, which acts just like your TEXT as the key.

Final thought is that `kwargs` is JSONB currently, which works well with the underlying Python dependencies, but I'd like to structure it as much as possible for final 3.0. We should find a way to flag this obviously as an alpha API, that will be broken and eventually dropped when a final version is available.

montanalow (changes requested):
You can add a migration to create the function for people who upgrade, e.g. 
https://github.com/postgresml/postgresml/blob/master/pgml-extension/sql/pgml--2.7.13--2.8.0.sql

ns1000:
So it turns out the batching is not really necessary to achieve speed. When running on CPU within the Postgres python VM, you really need to torch.set_num_threads(1) in order to get the maximum speed. Leaving it to the default value, which is the number of CPUs was creating the slow down problems for me. It will still use all the CPUs when use threads=1.

I am using a debian system, with python 3.11 and postgres 16 to test all this.

montanalow:
> So it turns out the batching is not really necessary to achieve speed. When running on CPU within the Postgres python VM, you really need to torch.set_num_threads(1) in order to get the maximum speed. Leaving it to the default value, which is the number of CPUs was creating the slow down problems for me. It will still use all the CPUs when use threads=1.
> 
> I am using a debian system, with python 3.11 and postgres 16 to test all this.

Ah, so this is actually another hit on #1161

PR #1201, opened 2023-11-28
Python & JavaScript CLI

levkk:
1. Add Python CLI.
2. Add JavaScript CLI in a separate package because npm doesn't allow a package to contain a module and a binary at the same time.
3. The CLI has just one command: `connect`, which creates a Postgres FDW connection from a PostgresML db to another PostgreSQL database.
4. Added `PGML_CHECKOUT_TIMEOUT` environment variable to control how long we wait for a connection in the SDK pool. Default: 5 seconds. This includes connect timeout for new pools.
5. `cargo fmt`.


montanalow:
Why not Rust? Binary distributable would make life easier.

levkk:
> Why not Rust? Binary distributable would make life easier.

Thought about it. The easiest way to distribute a CLI in the Rust world is with `cargo install`, which requires people to have the Rust toolchain installed. Conversely, if we use our SDK as the distribution vehicle, our CLI is now available to everyone who's using our SDK already in whatever language they choose.

montanalow:
Right, but we'd have to implement it in each language to accomplish what you're stating with this Python first approach. This method excludes Rust & JS, whereas we could follow the same Rust first method, and get Python/JS distribution for relatively free with the same behavior on all runtimes.

montanalow:
i.e. all SDKs should have a CLI, and the logic for the CLI should be implemented in the core Rust version.

levkk:
Take a look at the PR, the pieces are all there. The only reason I scoped this to Python is because I know how to make the module callable. In JS, I need to do some research.

PR #1206, opened 2023-11-29
Added OpenSourceAI and conversational support in the extension

SilasMarvin:


levkk:
🚀 🌔 🏆 

Quick question re: API change from String to JsonB for the transform_stream, is that backwards compatible? I don't support we have a lot of adoption of those new methods yet since we haven't actually released 2.8.0 yet "officially", but still would be nice to keep those working if we can.

SilasMarvin:
> 🚀 🌔 🏆
> 
> Quick question re: API change from String to JsonB for the transform_stream, is that backwards compatible? I don't suppose we have a lot of adoption of those new methods yet since we haven't actually released 2.8.0 yet "officially", but still would be nice to keep those working if we can.

It is a breaking change unfortunately. It was a bad choice on my side to make it a string. I spoke with Montana about it, and it hasn't been released and I'm the only one using it so not too worried. Before we ship and merge this I need to make adjustments to our frontend or it will break something though

levkk:
Sounds good!

pgml-extension/src/bindings/transformers/transform.rs
```diff
@@ -23,28 +23,30 @@ impl TransformStreamIterator {
...
-        Python::with_gil(|py| -> Result<Option<String>, PyErr> {
+        Python::with_gil(|py| -> Result<Option<JsonB>, PyErr> {
             let code = "next(python_iter)";
             let res: &PyAny = py.eval(code, Some(self.locals.as_ref(py)), None)?;
             if res.is_none() {
                 Ok(None)
             } else {
-                let res: String = res.extract()?;
-                Ok(Some(res))
+                eprintln!("\nHERE WE ARE!\n");
```
levkk, line 35: Universal debugger.

SilasMarvin: Hahah yes. Nice catch

PR #1211, opened 2023-12-05
bump sdk for dashboard

montanalow:


pgml-dashboard/Cargo.toml
```diff
@@ -29,7 +29,7 @@ log = "0.4"
 markdown = "1.0.0-alpha.14"
 num-traits = "0.2"
 once_cell = "1.18"
-pgml = { version = "0.9.4", path = "../pgml-sdks/pgml/" }
+pgml = { version = "0.9.6", path = "../pgml-sdks/pgml/" }
```
levkk, line 32: Pretty sure the SDK is at 0.10.0 now, worth a double check. Maybe we forgot to bump the Rust version.

PR #1250, opened 2023-12-20
Dan docs update

chillenberger:
Screenshot 2023-12-20 at 2 16 26 PM


montanalow:
For docs bad paths, it'd be nice if we could still render the left nav, since that'd be an easy recovery path when we move the docs around in gitbook.

chillenberger:
Does something like this work? 

Screenshot 2023-12-20 at 3 07 59 PM

PR #1251, opened 2023-12-20
should be ok not 400

chillenberger:
- Change test after changing behavior of docs bad path.

### DIFF
--- a/pgml-dashboard/src/api/cms.rs
+++ b/pgml-dashboard/src/api/cms.rs
@@ -550,7 +550,7 @@ This is the end of the markdown
         let rsp = req.dispatch().await;
 
         assert!(
-            rsp.status() == Status::NotFound,
+            rsp.status() == Status::Ok,
             "Returned status {:?}",
             rsp.status()
         );


pgml-dashboard/src/api/cms.rs
```diff
@@ -550,7 +550,7 @@ This is the end of the markdown
         let rsp = req.dispatch().await;
 
         assert!(
-            rsp.status() == Status::NotFound,
+            rsp.status() == Status::Ok,
```
levkk, line 553: Should be 404 when doc not found. If we return ok, google will index the not found page and real and pollute the search index.

chillenberger: Ok, I can switch it back.  We will have to wait on keeping the lefthand nav on bad path then.

levkk: I don't understand what the connection between this and the leftnav is with the current context.

chillenberger: After a bad path Montana would like to other docs there on the left to easy navigate away from the 404 response. 


Screenshot 2023-12-20 at 4 00 03 PM

levkk: Ah I see, you'll just need a custom 404 page for the docs instead of using the default one. You can set that up with a 404 catcher in rocket.

PR #1276, opened 2024-01-05
Activate venv at server start and remove the check from functions

SilasMarvin:


pgml-extension/src/lib.rs
```diff
@@ -24,6 +24,7 @@ extension_sql_file!("../sql/schema.sql", name = "schema");
 #[cfg(not(feature = "use_as_lib"))]
 #[pg_guard]
 pub extern "C" fn _PG_init() {
+    bindings::python::activate().expect("Error setting python venv");
```
levkk, line 27: This works? I ask because each connection forks from the postmaster. It gets the venv set correctly? :O

SilasMarvin: It works for me locally. I can test from multiple different connections.

SilasMarvin: Can confirm tested from three different connections and it worked

levkk: Well, that was easy.

levkk (approved):

montanalow (approved):
🎉

PR #1285, opened 2024-01-10
New site search

SilasMarvin:


pgml-sdks/pgml/src/collection.rs
```diff
@@ -718,6 +718,19 @@ impl Collection {
         }
     }
 
+    #[instrument(skip(self))]
```
levkk, line 721: The instrument macros are surprisingly expensive. If you benchmark it, I think it adds 30-40ms of latency per method execution. Maybe I did something wrong in my benchmarks, but when I removed the instrument call, everything went back to normal.

PR #1290, opened 2024-01-18
stratified random sampling

ChuckHend:
Implements a stratified random sampling strategy, and sets that as the new default for `test_sampling`.

### DIFF
--- a/pgml-sdks/pgml/javascript/examples/package-lock.json
+++ b/pgml-sdks/pgml/javascript/examples/package-lock.json
@@ -10,24 +10,27 @@
       "license": "ISC",
       "dependencies": {
         "dotenv": "^16.3.1",
-        "pgml": "^0.9.0"
+        "pgml": "^1.0.0"
       }
     },
     "node_modules/dotenv": {
-      "version": "16.3.1",
-      "resolved": "https://registry.npmjs.org/dotenv/-/dotenv-16.3.1.tgz",
-      "integrity": "sha512-IPzF4w4/Rd94bA9imS68tZBaYyBWSCE47V1RGuMrB94iyTOIEwRmVL2x/4An+6mETpLrKJ5hQkB8W4kFAadeIQ==",
+      "version": "16.4.5",
+      "resolved": "https://registry.npmjs.org/dotenv/-/dotenv-16.4.5.tgz",
+      "integrity": "sha512-ZmdL2rui+eB2YwhsWzjInR8LldtZHGDoQ1ugH85ppHKwpUHL7j7rN0Ti9NCnGiQbhaZ11FpR+7ao1dNsmduNUg==",
       "engines": {
         "node": ">=12"
       },
       "funding": {
-        "url": "https://github.com/motdotla/dotenv?sponsor=1"
+        "url": "https://dotenvx.com"
       }
     },
     "node_modules/pgml": {
-      "version": "0.9.1",
-      "resolved": "https://registry.npmjs.org/pgml/-/pgml-0.9.1.tgz",
-      "integrity": "sha512-CzXFegNtZEIcY8u+ZoBOmfzh2bYWGCWFt3aUZwu1dHcokzQ7mlGs4eIw1KijWv6ieKylFS33oHlh1uzqjkrSAg=="
+      "version": "1.0.0",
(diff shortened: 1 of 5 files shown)


pgml-extension/src/orm/snapshot.rs
```diff
@@ -742,26 +737,7 @@ impl Snapshot {
...
-                false => {
-                    if self.test_sampling == Sampling::random {
-                        "ORDER BY random()"
-                    } else {
-                        ""
-                    }
-                }
-            },
-        )
+        self.test_sampling.get_sql(&self.relation_name(), self.columns.clone())
```
montanalow, line 740: I think this will re-randomize or stratify if the snapshot has been materialized, which would make training runs not reproducible. I think we'll need something akin to the previous version for materialized runs, or this for dynamic runs.

pgml-extension/src/orm/sampling.rs
```diff
@@ -1,11 +1,14 @@
...
 use serde::Deserialize;
 
+use super::snapshot::Column;
+
 #[derive(PostgresEnum, Copy, Clone, Eq, PartialEq, Debug, Deserialize)]
 #[allow(non_camel_case_types)]
 pub enum Sampling {
     random,
     last,
+    stratified_random,
```
montanalow, line 11: Thinking ahead to other stratification strategies, that use columns other than the target y_column_name, to guarantee you have true out of sample rows. e.g. `user_id` may be used to train with multiple instances from a particular user, but you want to ensure there is no data leakage where the model is just memorizing user_ids rather than the more abstract, so user_id should be excluded as a feature, but used for stratification.

I think that could work as an additional `stratified_column_name`  parameter to train. In that case though, the sampling wouldn't be `stratified_random`. So we'd need to add a different stratified type, or we could just call this `stratified`, and if you don't specify a column, it's random by the y_column_name. If you do specify stratification column(s), then those column(s) get removed from features, and strictly used for stratification.

This can happen in a follow up PR, I'm just commenting so we get a forward looking name on the sampling strategy.

montanalow (commented):
Thanks for hammering this out!

pgml-extension/src/orm/snapshot.rs
ChuckHend: ah yes,  you're right. i think i can adjust this to account for whats in `self.materialized `

montanalow:
This will also need a migration in `sql/` to update the Postgres sampling enum.

ChuckHend:
> This will also need a migration in sql/ to update the Postgres sampling enum.

It looks like 2.8.2 hasn't gone out yet so I put the migration in `./sql/pgml--2.8.1--2.8.2.sql` . I can create a 2.8.3 migration if that would be preferred though.

pgml-extension/src/orm/sampling.rs
ChuckHend: Good ideas/points. I changed it to just `stratified`. For this PR, there will be no ability to specify the columns to stratify by, `stratified` only uses y_column_name. I think that should be a non-breaking change to add an optional parameter in the future that would change it from y_column_name to something else.

pgml-extension/sql/pgml--2.8.1--2.8.2.sql
```diff
@@ -25,3 +25,109 @@ CREATE  FUNCTION pgml."deploy"(
...
 ALTER TYPE pgml.strategy ADD VALUE 'specific';
+
+-- src/orm/sampling.rs:6
+-- pgml::orm::sampling::Sampling
+DROP TYPE IF EXISTS pgml.Sampling;
+CREATE TYPE pgml.Sampling AS ENUM (
+	'random',
+	'last',
+	'stratified'
+);
```
ChuckHend, line 36: I still need to do some more testing to make sure this migration works as intended.

@montanalow , are there integration tests that assert migrations work? I didn't see any...

montanalow: There aren't integration tests. I typically run the test/test.sql to populate a database with the previous version, then `later extension pgml update`.

montanalow: You'll want to `alter type add value...` here rather than dropping the enum, as that would only work on an empty database.

PR #1294, opened 2024-01-19
Dan blog landing page

chillenberger:


pgml-dashboard/src/api/cms.rs
```diff
@@ -61,35 +62,66 @@ lazy_static! {
...
-    pub html: String,
+    pub contents: String,
+    pub doc_type: Option<DocType>,
 }
 
+// Gets document markdown
 impl Document {
     pub async fn from_path(path: &PathBuf) -> anyhow::Result<Document, std::io::Error> {
         warn!("path: {:?}", path);
+
```
levkk, line 106: Never noticed the ^ `warn!` above. It's a bit too noisy for production, could we move that to `debug!`?

chillenberger: yeah, I can do that in my next merge.

PR #1300, opened 2024-01-23
fix doctype

montanalow:


montanalow:
@chillenberger I see your PR around this too, thought it'd be able to fix than report the missing `/` in both the regex and url.

chillenberger:
I was thinking about just using the tools from Pathbuf to determine if it is doc, blog, or careers.  
`let doc_type = match path.strip_prefix(config::cms_dir()) {
            Ok(path) => {
                match path.into_iter().next() {
                    Some(dir) => {
                        match &PathBuf::from(dir).display().to_string()[..] {
                            "blog" => Some(DocType::Blog),
                            "docs" => Some(DocType::Docs),
                            "careers" => Some(DocType::Careers),
                            _ => None
                        }
                    },
                    _ => None
                } 
            }, 
            _ => None,
        };`

montanalow:
Could also pass in the collection to avoid having to parse at all. Or make
this a collection method.

chillenberger:
It felt too specific to the doc to put in collection and I wanted to minimize the amount of info the caller needed since I wasn't sure what Silas would have access too.  Here is what I am thinking https://github.com/postgresml/postgresml/pull/1301

PR #1310, opened 2024-02-01
Install huggingface transformers library

silverjam:
This library seems to be required for certain models that require code to be run.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -9,11 +9,9 @@ RUN apt update && \
 		coreutils \
 		sudo \
 		openssl \
-		git \
-		build-essential \
 		python3-pip \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -8,7 +8,10 @@ RUN apt update && \
 		gnupg \
 		coreutils \
 		sudo \
-		openssl
+		openssl \
+		python3-pip \
+		&& \
+	pip install transformers
```
levkk, line 14: should  be in requirements.txt

montanalow:
This is also documented as a required step in a few places, e.g. https://postgresml.org/docs/resources/developer-docs/installation#python-dependencies

PR #1314, opened 2024-02-06
Fixed some queries

SilasMarvin:


pgml-sdks/pgml/src/collection.rs
```diff
@@ -936,6 +937,8 @@ impl Collection {
                 .fetch_all(&pool)
                 .await;
 
+                println!("\n\n{result:?}\n\n");
```
levkk, line 940: Remove this, it'll spam our users logs.

pgml-sdks/pgml/src/lib.rs
```diff
@@ -354,18 +354,21 @@ mod tests {
...
         collection
-            .upsert_documents(generate_dummy_documents(3))
+            .upsert_documents(generate_dummy_documents(5))
             .await?;
         let results = collection
             .vector_search("Here is some query", &mut pipeline, None, None)
             .await?;
-        assert!(results.len() == 3);
+
+        println!("\n{results:?}\n");
```
levkk, line 369: Same.

pgml-sdks/pgml/src/collection.rs
SilasMarvin: Oh wow great catches sorry

pgml-sdks/pgml/src/queries.rs
```diff
@@ -207,31 +214,20 @@ FROM
...
-  INNER JOIN %s chunks ON chunks.id = comparison.chunk_id 
-  INNER JOIN %s documents ON documents.id = chunks.document_id 
-  ORDER BY 
-  comparison.score DESC 
-  LIMIT 
+  %s embeddings
+INNER JOIN %s chunks ON chunks.id = embeddings.chunk_id 
+INNER JOIN %s documents ON documents.id = chunks.document_id 
+ORDER BY 
+  embeddings.embedding <=> $1::vector ASC
```
levkk, line 229: This uses the index but does it mean the same? The sort on the right sorts by highest scores first, while this scores by lowest scores first. In fact, won't this give the worst results by similarity?

SilasMarvin: Yes it means the same. <=> is the cosine distance function. We want ones with the lowest distance. 1 - <=> is the cosine similarity. You want ones with high similarity which is the same as low distance

pgml-sdks/pgml/src/queries.rs
```diff
@@ -134,8 +134,10 @@ WHERE
   AND id NOT IN (
```
montanalow, line 134: Exists is more idiomatic
```suggestion
  AND NOT EXISTS (
```

pgml-sdks/pgml/src/queries.rs
```diff
@@ -157,8 +159,10 @@ WHERE
...
```
montanalow, line 159:
```suggestion
  AND NOT EXISTS (
```

pgml-sdks/pgml/src/queries.rs
montanalow: The query planner should optimize this as equivalent, but you could also do a `LEFT OUTER JOIN "table" ON chunk_id = ... WHERE "table".id IS NULL` which will select only the missing rows from "table", if the subquery shows up as an expensive component.

PR #1317, opened 2024-02-16
Silas new site search and SDK 1.0 release

SilasMarvin:


montanalow (approved):

pgml-sdks/pgml/src/migrations/pgml--0.9.2--1.0.0.rs
```diff
@@ -0,0 +1,9 @@
+use sqlx::PgPool;
+use tracing::instrument;
+
+#[instrument(skip(_pool))]
+pub async fn migrate(_pool: PgPool, _: Vec<i64>) -> anyhow::Result<String> {
+    anyhow::bail!(
+        "There is no automatic migration to SDK version 1.0. Please just upgrade the SDK and create a new collection",
```
levkk, line 7: I'm sure we can migrate people. Asking people to start anew isn't something we should be doing. If we can't write an automatic migration now, that's fine. In this case, we can ask them to reach out on Discord and we can help them migrate.

SilasMarvin: We can write a migration script although it is not trivial and we do need to make some decisions on what indexes to give them. This was @montanalow 's recommendation. If we want a migration script I can try and write one.

montanalow: Just a note on copy, I try not to use the word "just..." as it may seem to trivialize efforts. We could also include a "Please upgrade the SDK and create a new collection, or contact your PostgresML support to create a migration plan."

SilasMarvin: Updated! https://github.com/postgresml/postgresml/pull/1317/commits/28cd41f3b66694924526fdc8bf5e5e06a0f184f6

levkk (approved):
:rocket:

aplchian:
nice

PR #1320, opened 2024-02-17
Dan doc landing page

chillenberger:
- Create Docs landing page
- New Docs article page
- Change look of scrollbar 
- Remove index from careers and blogs
- Redesigen TOC


montanalow:
Now that we're indexing all docs/blogs premptively, is it easy to build the index page urls without the trailing `/`? The reason I added that `/` in the first place was to prevent having to check on disk if it was a file or directory before appending the `README.md` name. If we can avoid the extra check now, it'd be nice to be consistent.

chillenberger:
> Now that we're indexing all docs/blogs premptively, is it easy to build the index page urls without the trailing `/`? The reason I added that `/` in the first place was to prevent having to check on disk if it was a file or directory before appending the `README.md` name. If we can avoid the extra check now, it'd be nice to be consistent.

When the user navigates to a doc url we still use the url path to build the file path in the get_content_path method.  We can't get away from the check now.  We could do what your asking, but it's not trivial enough to throw into this PR.

PR #1341, opened 2024-03-01
fix shared_preload_libraries in extension update test

ChuckHend:


montanalow:
I spent some time cleaning up our CI in #1340 and picked up this change as well. Thanks!

PR #1352, opened 2024-03-04
Added the-1.0-sdk-is-here blog post

SilasMarvin:


pgml-dashboard/src/api/cms.rs
```diff
@@ -26,6 +26,7 @@ lazy_static! {
         "Blog",
         true,
         HashMap::from([
+            ("the-1.0-sdk-is-here", "the-1.0-sdk-is-here"),
```
montanalow, line 29: You only need this if you're mapping one url to another to fix renamed posts. This looks identical?

PR #1360, opened 2024-03-06
Refactor the initialization of GUC parameters.

higuoxing:
Managing GUC parameters in different places is hard to maintain. This patch organizes GUC definitions in a single place. Also, we use define_xxx_guc() APIs to define these parameters and it will allow us to manage GucContext, GucFlags in future.

P.S., the test case test_trusted_model doesn't seem correct. I fixed it in this patch.


pgml-extension/src/bindings/transformers/whitelist.rs
```diff
@@ -145,18 +152,18 @@ mod tests {
...
+        set_config(PGML_HF_TRUST_WHITELIST.0, model).unwrap();
 
         let task_json = format!(json_template!(), model, false);
         let task: Value = serde_json::from_str(&task_json).unwrap();
         assert!(verify_task(&task).is_ok());
 
         let task_json = format!(json_template!(), model, true);
         let task: Value = serde_json::from_str(&task_json).unwrap();
-        assert!(verify_task(&task).is_ok());
+        assert!(verify_task(&task).is_err());
```
higuoxing, line 164: In the initial commit of this commit, this step should fails with `RemoteCodeNotTrusted`.

https://github.com/postgresml/postgresml/blob/f4e87c5b85013ec74ec583e33343745396320442/pgml-extension/src/bindings/transformers/whitelist.rs#L189

pgml-extension/src/config.rs
```diff
@@ -1,16 +1,55 @@
...
+        &PGML_HF_WHITELIST.1,
+        GucContext::Userset,
+        GucFlags::default(),
+    );
+    GucRegistry::define_bool_guc(
+        PGML_HF_TRUST_REMOTE_CODE.0,
+        "Whether model can execute remote codes",
+        "",
+        &PGML_HF_TRUST_REMOTE_CODE.1,
+        GucContext::Userset,
```
higuoxing, line 42: Shall we make these GUCs `SUSET`?

>  SUSET options can be set at postmaster startup, with the SIGHUP mechanism, or from the startup packet or SQL if you're a superuser.

https://github.com/postgres/postgres/blob/d93627bcbe5001750e7611f0e637200e2d81dcff/src/include/utils/guc.h#L74

higuoxing:
Merged in commit https://github.com/postgresml/postgresml/commit/f674f70b542b164b98a9a63e009b2b3b7a73afdb

PR #1364, opened 2024-03-12
Add support for mixedbread embeddings

kczimm:
Bumps sentence-transformers version to support mixedbread-ai/mxbai-embed-2d-large-v1


kczimm:
Fixed in #1365

PR #1386, opened 2024-03-22
Added once_cell crate to make RUNTIME lazy

SilasMarvin:


pgml-sdks/pgml/src/lib.rs
```diff
@@ -123,24 +124,15 @@ fn internal_init_logger(level: Option<String>, format: Option<String>) -> anyhow
...
+static RUNTIME: Lazy<Runtime> = Lazy::new(|| {
+    Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .expect("Error creating tokio runtime")
+});
 
-#[allow(dead_code)]
 fn get_or_set_runtime<'a>() -> &'a Runtime {
-    unsafe {
```
levkk, line 131: Oh wow, makes sense. This is totally not thread safe!

PR #1387, opened 2024-03-23
Added vector_search_local

SilasMarvin:


montanalow (commented):
What's the use case here?

SilasMarvin:
> What's the use case here?

Calling from Rust when variables aren't mutable. For instance, we use the document_search_local in our site. We only need the collection and pipeline to be mutable if we don't know the embeddings are done locally. If the embeddings aren't done locally the query will fail, we update the collection and pipeline with the correct data, and then retry.

montanalow:
Got it. So, local in this case actually refers to the database, not to the client, as opposed to foreign embeddings from OpenAI, that require a mutable pipeline to implement the retry logic. And we need to query the database find out the local/foreign status of a pipeline, but we're optimistically trying to do it all as a single query. Makes sense.

I think I would've called this native/unified/embedded/inside rather than local, but I get it now.

PR #1393, opened 2024-04-03
Onboarding components fixes

levkk:
### New components
1. Gray heading.
2. Primary card that doesn't break absolute-positioned elements (dropdown) rendered in other cards.

### Bug fixes
1. Radio input correctly notifies caller in case of change.
2. Small table shouldn't have a hover effect.

### Small additions
1. Added optional anchor to `SwitchV2`. Helps with navigation using Turbo frames or streams.
3. A few helpful setters.
4. Split up unit displayed in the input and unit displayed in the cost estimate in `RangeGroupV2`
5. More constructors for `StimulusAction` to avoid importing enum.


pgml-dashboard/src/components/cards/rgb/mod.rs
```diff
@@ -19,20 +25,44 @@ impl Rgb {
...
+
+    pub fn link_action(mut self, action: StimulusAction) -> Self {
+        self.link_action = action.into();
+        self
+    }
+
+    pub fn h_100(mut self) -> Self {
+        self.controller_classes.push("h-100".into());
+        self.card_classes.push("h-100".into());
+        self
```
chillenberger, line 64: for something like this is it possible to just always have the h-100 on and for instances where you need to limit the height, place it in a wrapper that does not have h-100?  Just brainstorming, this could turn into a slippery slope of having a function for every specific situation.  I think if it is possible to solve problems like this with wrappers, we should.

pgml-dashboard/src/components/cards/primary/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+        Primary {
+            component,
+            style: "".into(),
+        }
+    }
+
+    pub fn z_index(mut self, index: i64) -> Self {
+        self.style = format!("position: relative; z-index: {};", index);
+        self
+    }
```
chillenberger, line 22: Same thing as the next comment, is it possible to solve this with a wrapper?

pgml-dashboard/src/components/headings/gray/mod.rs
```diff
@@ -0,0 +1,18 @@
+use pgml_components::component;
+use sailfish::TemplateOnce;
+
+#[derive(TemplateOnce, Default)]
+#[template(path = "headings/gray/template.html")]
+pub struct Gray {
+    value: String,
+}
+
+impl Gray {
```
chillenberger, line 10: If we want to go down this road, maybe one heading that takes a color, or has a setter function for the desired color, rather than a new component for every header.

pgml-dashboard/src/components/headings/gray/gray.scss
```diff
@@ -0,0 +1,3 @@
+span[data-controller="headings-gray"] {
+    color: #{$gray-400};
+}
```
chillenberger, line 3: You could add color = gray-400 to the typography.scss file and just use the class here.  Utility classes for the gray scale is always nice.

pgml-dashboard/src/components/cards/rgb/mod.rs
levkk: By default, the card occupies as much space as it needs to render its contents. The `.h-100` here helps cards look uniform when placed inside a grid. Our cards don't have the same contents, so their heights will vary when placed inside a grid like `.row` or `.d-flex`. If we have `.h-100` by default, the card will always occupy the full height of its container which is not desired when the card is used by itself inside something like a `<div class="container">`. So `.h-100` is a special case that we just happen to use quite often, but not always.

A wrapper can't make its child occupy 100% of its space unless we add additional CSS to the wrapper knowing that the child needs to do this. For example, here we could do something like:

```sass
.card-wrapper {
    .card {
          @extend .h-100;
    }
}
```

which now that I'm writing this isn't too bad. One issue with this is more of a theoretical technicality, where a component "reaches inside" another component and modifies its state which could cause weird bugs later on. It's typically best to have the component use public methods to modify its state in a predicable  way.

Maybe if we name this slightly differently, it could be better? I'm not leaning strongly one way or the other.

pgml-dashboard/src/components/cards/primary/mod.rs
levkk: Same issue as above. The wrapper would have to modify the style of its child.

pgml-dashboard/src/components/headings/gray/mod.rs
levkk: Yeah I agree, these are too small to be individual components.

PR #1453, opened 2024-05-10
RAG in the SDK

SilasMarvin:


pgml-sdks/pgml/src/lib.rs
```diff
@@ -1959,4 +2152,630 @@ mod tests {
...
+                &mut pipeline,
+            )
+            .await?;
+        assert!(!results["rag"].as_array().unwrap()[0]
+            .as_str()
+            .unwrap()
+            .is_empty());
+
+        // Multi-variable test
+        let results = collection
```
SilasMarvin, line 2232: @montanalow This is an example of the kind of query you can do

pgml-sdks/pgml/javascript/tests/typescript-tests/test.ts
```diff
@@ -92,29 +92,30 @@ it("can search", async () => {
       query: {
         full_text_search: { body: { query: "Test", boost: 1.2 } },
         semantic_search: {
-          title: { query: "This is a test", boost: 2.0 },
```
montanalow, line 95: Not for this PR, but in the future it might be nice to have a "needle" document, and use it's ID, so you don't have unstable result IDs because all the documents are equally similar "This is a test document".

montanalow (approved):

PR #1459, opened 2024-05-13
Add migration and bump extension version

SilasMarvin:


kczimm (approved):

levkk (approved):
Neat. Breaking change technically but since that API was broken already, is it though? I don't think so!

PR #1471, opened 2024-05-17
A few Python fixes

levkk:
1. Show the Python executable when creating the extension (validating Python dependencies).
2. Print virtualenv error to stderr instead of stdout; this way, it will be captured in Postgres logs.
3. Add a working Python 3.12 environment, with support for `pgml.train()`, `pgml.embed()`, `pgml.transform()`, `pgml.chunk()`.


kczimm (approved):
LGTM

pgml-extension/src/bindings/python/python.py
```diff
@@ -21,7 +21,7 @@ def activate_venv(venv):
         __venv = venv
         return True
     else:
-        print("Virtualenv not found: %s" % venv)
+        print("virutalenv not found: %s" % venv, file=sys.stderr)
```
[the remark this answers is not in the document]
kczimm: Oh, sorry about the indent. It was just the spelling error with `virtualenv`.

pgml-extension/requirements.py312.txt
```diff
@@ -0,0 +1,97 @@
+accelerate==0.30.1
```
montanalow, line 1: I'm guessing these are linux python 3.12 requirements. Why not replace requirements.linux.txt with this, and document 3.12 as the Python version required, to avoid sprawling too much. These frozen requirements are for deployments/dockerfiles more than development.

levkk: > I'm guessing these are linux python 3.12 requirements.
Don't have to guess, it's in the PR description.

> Why not replace requirements.linux.txt with this, and document 3.12 as the Python version required, to avoid sprawling too much.
`ray` doesn't run on Python 3.12 and above, which is required by `vllm` and `pgml.transform_stream()`. So we actually don't have full support for our extension in Python 3.12.

> These frozen requirements are for deployments/dockerfiles more than development.
Well, we have to start somewhere. I need this to run on Python 3.12 locally since I've seen the light and now am using Arch with latest available Python version. Arch users now officially outnumber Ubuntu users, so ...there you go.

PR #1475, opened 2024-05-22
Embeddings support in the SDK

levkk:
### SDK
1. Added `pgml.embed()` support into `Builtins`, with single and batch mode.

### DIFF
--- a/pgml-sdks/pgml/src/builtins.rs
+++ b/pgml-sdks/pgml/src/builtins.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use rust_bridge::{alias, alias_methods};
 use sqlx::Row;
 use tracing::instrument;
@@ -13,7 +14,7 @@ use crate::{get_or_initialize_pool, query_runner::QueryRunner, types::Json};
 #[cfg(feature = "python")]
 use crate::{query_runner::QueryRunnerPython, types::JsonPython};
 
-#[alias_methods(new, query, transform, embed)]
+#[alias_methods(new, query, transform, embed, embed_batch)]
 impl Builtins {
     pub fn new(database_url: Option<String>) -> Self {
         Self { database_url }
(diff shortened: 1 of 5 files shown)


pgml-sdks/pgml/src/builtins.rs
```diff
@@ -87,6 +88,55 @@ impl Builtins {
...
+    pub async fn embed_batch(&self, model: &str, texts: Json) -> anyhow::Result<Json> {
+        let texts = texts
+            .0
+            .as_array()
+            .with_context(|| "embed_batch takes an array of texts")?
+            .into_iter()
+            .map(|v| {
+                v.as_str()
+                    .with_context(|| "only text embeddings are supported")
+                    .unwrap()
```
kczimm, line 124: You can collect into a `Result<Vec<_>>` so you can `?` the `Result` instead of panicking if the value is not a `str`.

kczimm: Something like

```rust
let texts = texts
            .0
            .as_array()
            .with_context(|| "embed_batch takes an array of texts")?
            .iter()
            .map(|v| {
                v.as_str()
                    .with_context(|| "only text embeddings are supported")
                    .map(|s| s.to_string())
            })
            .collect::<anyhow::Result<Vec<String>>>()?;
```

I think will work.

SilasMarvin:
A test for both would be nice as well

pgml-sdks/pgml/src/builtins.rs
levkk: I don't think we want that. The API accepts an array of strings, so if someone passes in an array of objects that can be cast to a string, they will get embeddings of integers, or whatever else they passed in, which will make them think that everything is fine.

levkk:
> A test for both in Rust, Python, and JavaScript would be nice

Do these tests run in CI or only locally? They require a PostgresML installation to be running somewhere right?

levkk:
I can't seem to find instructions for setting up the JS environment for testing this. I did the Python one from memory, but it would be nice to have a short README for developers. I'm going to skip the JS tests for now and come back to them when I have a moment.

pgml-sdks/pgml/src/builtins.rs
kczimm: We are still calling `as_str` here so if it was an object, it wouldn't be treated as a string. I was merely suggesting using `Result::Err` instead of `unwrap`. But all is good.

PR #1498, opened 2024-06-03
Silas add ranking

SilasMarvin:
This integration brings reranking into postgresml.

```
pgml=# SELECT pgml.rank('mixedbread-ai/mxbai-rerank-large-v1', 'test', array_agg(md5(random()::text)), '{"return_documents": false, "top_k": 10}') FROM generate_series(1, 100);
           rank            
---------------------------
 (58,0.20096051692962646,)
 (91,0.2007983922958374,)
 (84,0.1950932741165161,)
 (83,0.1925133764743805,)
 (7,0.1918289214372635,)
 (15,0.1851881593465805,)
 (67,0.18225009739398956,)
 (94,0.1795625537633896,)
 (40,0.17863182723522186,)
 (97,0.177006796002388,)
(10 rows)
```

### DIFF
--- a/pgml-extension/src/bindings/transformers/transformers.py
+++ b/pgml-extension/src/bindings/transformers/transformers.py
@@ -12,7 +12,7 @@ import numpy
 import orjson
 from rouge import Rouge
 from sacrebleu.metrics import BLEU
-from sentence_transformers import SentenceTransformer
+from sentence_transformers import SentenceTransformer, CrossEncoder
 from sklearn.metrics import (
     mean_squared_error,
     r2_score,
@@ -498,6 +498,33 @@ def transform(task, args, inputs, stream=False):
     if stream:
         return pipe.stream(inputs, **args)
     return orjson.dumps(pipe(inputs, **args), default=orjson_default).decode()
+
+
+def create_cross_encoder(transformer):
+    return CrossEncoder(transformer)
+
+
+def rank_using(model, query, documents, kwargs):
+    if isinstance(kwargs, str):
+        kwargs = orjson.loads(kwargs)
+
+    # The score is a numpy float32 before we convert it
+    return [
+        {"score": x.pop("score").item(), **x}
+        for x in model.rank(query, documents, **kwargs)
+    ]
+
+
+def rank(transformer, query, documents, kwargs):
+    kwargs = orjson.loads(kwargs)
+
+    if transformer not in __cache_sentence_transformer_by_name:
+        __cache_sentence_transformer_by_name[transformer] = create_cross_encoder(
+            transformer
+        )
+    model = __cache_sentence_transformer_by_name[transformer]
+
+    return rank_using(model, query, documents, kwargs)
 
 
 def create_embedding(transformer):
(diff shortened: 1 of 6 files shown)


pgml-extension/src/bindings/transformers/transformers.py
```diff
@@ -500,6 +500,33 @@ def transform(task, args, inputs, stream=False):
...
+        for x in model.rank(query, documents, **kwargs)
+    ]
+
+
+def rank(transformer, query, documents, kwargs):
+    kwargs = orjson.loads(kwargs)
+
+    if transformer not in __cache_sentence_transformer_by_name:
+        __cache_sentence_transformer_by_name[transformer] = create_cross_encoder(
+            transformer
```
kczimm, line 523: If you pass kwargs through to `create_cross_encoder` we can specify the device, https://www.sbert.net/docs/package_reference/cross_encoder/cross_encoder.html?highlight=crossencoder#sentence_transformers.cross_encoder.CrossEncoder

We should do this for the `SentenceTransformer` constructor, too.

kczimm (commented):
We should bump the version (probably to 2.9.0 since it's an API addition)  and add the following migration file `sql/pgml--2.8.5--2.9.0.sql`:

```sql
-- src/api.rs:613
-- pgml::api::rank
CREATE  FUNCTION pgml."rank"(
	"transformer" TEXT, /* &str */
	"query" TEXT, /* &str */
	"documents" TEXT[], /* alloc::vec::Vec<&str> */
	"kwargs" jsonb DEFAULT '{}' /* pgrx::datum::json::JsonB */
) RETURNS TABLE (
	"corpus_id" bigint,  /* i64 */
	"score" double precision,  /* f64 */
	"text" TEXT  /* core::option::Option<alloc::string::String> */
)
IMMUTABLE STRICT PARALLEL SAFE 
LANGUAGE c /* Rust */
AS 'MODULE_PATHNAME', 'rank_wrapper';
```

pgml-extension/src/bindings/transformers/transformers.py
SilasMarvin: We will have to create a separate argument for this, or pop specific arguments from kwargs. If just pass kwargs straight through we will get an unexpected keyword argument error.

SilasMarvin:
> We should bump the version (probably to 2.9.0 since it's an API addition) and add the following migration file `sql/pgml--2.8.5--2.9.0.sql`:
> 
> ```sql
> -- src/api.rs:613
> -- pgml::api::rank
> CREATE  FUNCTION pgml."rank"(
> 	"transformer" TEXT, /* &str */
> 	"query" TEXT, /* &str */
> 	"documents" TEXT[], /* alloc::vec::Vec<&str> */
> 	"kwargs" jsonb DEFAULT '{}' /* pgrx::datum::json::JsonB */
> ) RETURNS TABLE (
> 	"corpus_id" bigint,  /* i64 */
> 	"score" double precision,  /* f64 */
> 	"text" TEXT  /* core::option::Option<alloc::string::String> */
> )
> IMMUTABLE STRICT PARALLEL SAFE 
> LANGUAGE c /* Rust */
> AS 'MODULE_PATHNAME', 'rank_wrapper';
> ```

Done! https://github.com/postgresml/postgresml/pull/1498/commits/745b190fbfe6b5a0551b8aeab46644c29caa5db0

kczimm (approved):

PR #1499, opened 2024-06-03
Allow to inject arbitrary components into the <head>

levkk:
Allow to inject any HTML into any page rendered by the open source code on any page on postgresml.org. Notable limitation currently is we can't control the `<head>` on `/docs` since the handler is completely inside the dashboard.

### DIFF
--- a/pgml-dashboard/src/components/layouts/head/mod.rs
+++ b/pgml-dashboard/src/components/layouts/head/mod.rs
@@ -1,4 +1,4 @@
-use pgml_components::component;
+use pgml_components::{component, Component};
 use sailfish::TemplateOnce;
 
 #[derive(TemplateOnce, Default, Clone)]
@@ -10,6 +10,7 @@ pub struct Head {
     pub preloads: Vec<String>,
     pub context: Option<String>,
     pub canonical: Option<String>,
+    pub additional_components: Vec<Component>,
 }
 
 impl Head {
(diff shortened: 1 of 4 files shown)


chillenberger:
Have you tried using this?  I have a suspicion you may end up with a bunch of head components that could cause bugs since "During rendering, Turbo Drive replaces the contents of the <body> element and merges the contents of the <head> element." So we just keep adding components to head without ever removing them and the components that are present depends on how the user moves through our site.

levkk:
Yeah it's in production right now for Hotjar.

chillenberger:
But hotjar should be on every page, correct? Imagine I add a component to scroll the page to the top, then on a different page I add a component to scroll to the bottom, both will exist in the head until a refresh occurs since the head is merged.

levkk:
Yeah I guess this is meant to inject HTML on every page. I thought Turbo updates head if it changes, but maybe not.

chillenberger:
It does update head, but it merges the existing head and the new head together. https://turbo.hotwired.dev/handbook/introduction#turbo-drive%3A-navigate-within-a-persistent-process we have hit some hard to diagnose bugs with this before.

levkk:
I see. Okay makes sense, this feature could be a bit dangerous, got it.

chillenberger:
Do we want hotjar on every page.  As of now if the user goes to postgresml.org then to postgresml.org/blog hotjar is in the header, but if the user goes directly to postgresml.org/blog, hotjar is not present.  We can add hotjar to cloud2 header component to fix this.

levkk:
We do want it on every page, but I wanted to avoid to add it to the open source code, since we don't want our open source users to accidentally load the hotjar code.

chillenberger:
If you add it to the cloud2 extend_header, open source will never see it, but it will be on every page in the production site. putting it here should be fine https://github.com/HyperparamAI/hypercloud/blob/71bd5877c6e1f87b3d59afd488ff42ecfee58960/cloud2/src/components/layouts/extend_head/template.html#L35

levkk:
Ah TIL, thanks. That solves my issue.

PR #1516, opened 2024-06-07
SDK - Added re-ranking into vector search

SilasMarvin:


pgml-sdks/pgml/src/lib.rs
```diff
@@ -1553,6 +1553,88 @@ mod tests {
...
+                            },
+                            "body": {
+                                "query": "Test document: 2",
+                                "parameters": {
+                                    "prompt": "passage: "
+                                },
+                                "boost": 1.0
+                            },
+                        }
+                    },
```
SilasMarvin, line 1616: @montanalow How does this "rerank" key look?

pgml-sdks/pgml/src/lib.rs
```diff
@@ -1553,6 +1553,88 @@ mod tests {
...
+                    "rerank": {
+                        "query": "Test document 2",
```
montanalow, line 1618: Seems like `query` is being repeated a few places in this example, which may be pretty typical. One enhancement would be to move the query string out and reuse it everywhere, and make passing specific sub clause query strings optional. Not a launch blocker though.

SilasMarvin: Got it, I will think more on making that  optional and reusing it, but will merge this and get it out in the meantime.

pgml-sdks/pgml/src/lib.rs
```diff
@@ -1553,6 +1553,88 @@ mod tests {
...
+                        "model": "mixedbread-ai/mxbai-rerank-base-v1",
+                        "num_documents_to_rerank": 100
```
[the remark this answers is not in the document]
SilasMarvin: Oh sorry missed this before merging. I think it might be a little confusing if we make it limit as we already have a limit key, and this isn't actually the limit. We already defined limit with llama index to mean the final number of items returned, but I'm not sure if they or langchain use it elsewhere.

PR #1528, opened 2024-06-14
make session backwards compatible

chillenberger:
- [x] make session cookie compatible with old cookie and recover no matter what
 - [x] use serde json to string
 - [x] add tests

### DIFF
--- a/pgml-dashboard/src/lib.rs
+++ b/pgml-dashboard/src/lib.rs
@@ -528,9 +528,8 @@ mod test {
     use super::*;
     use crate::components::sections::footers::MarketingFooter;
     use crate::guards::Cluster;
-    use crate::utils::config;
     use rocket::fairing::AdHoc;
-    use rocket::http::{Cookie, Status};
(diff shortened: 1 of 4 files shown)


pgml-dashboard/src/lib.rs
```diff
@@ -519,3 +522,176 @@ pub fn routes() -> Vec<Route> {
...
+            },
+            NotificationCookie {
+                id: "2".to_string(),
+                time_viewed: None,
+                time_modal_viewed: None,
+            },
+        ];
+
+        let response = client
+            .get("/notifications/product/modal/remove_modal?id=1")
```
levkk, line 554: This feels like the wrong place to implement this. This should be in our web app, and we should pass it through the context into the dashboard.

chillenberger: Not sure what you mean here.  The test, or the endpoint should be in web app? I felt like both are needed here since notifications include marketing notifications.

PR #1529, opened 2024-06-17
Session cookie deserializaiton

levkk:
1. Use serde for de(serialization).
2. Save the session cookie forever instead of just for 4 weeks.


levkk:
Merged into Dan's PR.

PR #1535, opened 2024-06-17
Added blog post semantic search in postgres in 15 minutes

SilasMarvin:


pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+
+June 15, 2024
+
+## What is and is not semantic search
+
+Semantic search uses machine learning to understand the meaning of text by converting it into numerical vectors, allowing for more accurate and context-aware search results.
+
+When users are unsure of the exact terms to search for, semantic search can uncover relevant information that traditional keyword searches might miss. This capability is particularly valuable for discovering content based on the intent and context of the search query, rather than relying solely on precise word matches.
+
+It is not a complete replacement for full-text search. In many cases, full-text search can outperform semantic search. Specifically, if a user knows the exact phrase they want to match in a document, full-text search is faster and guaranteed to return the correct result, whereas semantic search is only likely to return the correct result. Full-text search and semantic search can be combined to create powerful and robust search systems.
```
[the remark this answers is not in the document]
montanalow: I'd also go a bit further in this section to mention re-ranking is applicable to both, and as long as you need keyword search, you need classical support for entity recognition, and graph associations, which can also improve semantic matching.

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+
+Semantic search is not just for machine learning engineers. The system behind semantic search is relatively easy to implement, and thanks to new Postgres extensions like _pgml_ and _pgvector_, it is readily available to SQL developers. Just as modern SQL developers are expected to be familiar with and capable of implementing full-text search, they will soon be expected to implement semantic search as well.
```
montanalow, line 28:
```suggestion
Semantic search is not just for machine learning engineers. The system behind semantic search is relatively easy to implement, and thanks to new Postgres extensions like `pgml` and `pgvector`, it is readily available to SQL developers. Just as modern SQL developers are expected to be familiar with and capable of implementing full-text search, they will soon be expected to implement semantic search as well.
```

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+
+## Embeddings 101
+
+Semantic search is powered by embeddings. To understand how semantic search works, we must have a basic understanding of embeddings. 
+
+Embeddings are vectors. Given some text and some embedding model, we can convert text to vectors:
```
montanalow, line 34: more programmers are familiar with arrays than vectors if you're looking for a common explanation.

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+```
+
+!!!
+
+!!!
+
+We used the [pgml.embed](/docs/api/sql-extension/pgml.embed) PostresML function to generate an embedding of the sentence "Generating embeddings in Postgres is fun!" using the [mixedbread-ai/mxbai-embed-large-v1](https://huggingface.co/mixedbread-ai/mxbai-embed-large-v1) model from mixedbread.ai.
+
+The output size of the vector varies per model, and in _mxbai-embed-large-v1_ outputs vectors with 1024 dimensions: each vector contains 1024 floating point numbers.
```
montanalow, line 58:
```suggestion
The output size of the vector varies per model, and in `mxbai-embed-large-v1` outputs vectors with 1024 dimensions: each vector contains 1024 floating point numbers. 
```

I think we should use backticks for all identifiers, rather than italics.

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+
+!!!
+
+We used the [pgml.embed](/docs/api/sql-extension/pgml.embed) PostresML function to generate an embedding of the sentence "Generating embeddings in Postgres is fun!" using the [mixedbread-ai/mxbai-embed-large-v1](https://huggingface.co/mixedbread-ai/mxbai-embed-large-v1) model from mixedbread.ai.
+
+The output size of the vector varies per model, and in _mxbai-embed-large-v1_ outputs vectors with 1024 dimensions: each vector contains 1024 floating point numbers. 
+
+The vector this model outputs is not random. It is designed to capture the semantic meaning of the text. What this really means, is that sentences which are closer together in meaning will be closer together in vector space. 
+
+Let’s look at a more simple example. Let's assume we have a model called _simple-embedding-model_, and it outputs vectors with only 2 dimensions. Let’s embed the following three phrases: "I like Postgres", "I like SQL" and "Rust is the best":
```
montanalow, line 62:
```suggestion
Let’s look at a more simple example. Let's assume we have a model called `simple-embedding-model`, and it outputs vectors with only 2 dimensions. Let’s embed the following three phrases: "I like Postgres", "I like SQL" and "Rust is the best": 
```

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+embedding for 'Rust is the best'
+---------
+[-0.8, -0.9]
+```
+
+!!!
+
+!!!
+
+You'll notice how similar the vectors produced by the text "I like Postgres" and "I like SQL" are compared to "Rust is the best".  This is a artificial example, but the same idea holds true when translating to real models like _mixedbread-ai/mxbai-embed-large-v1_.
```
montanalow, line 98:
```suggestion
You'll notice how similar the vectors produced by the text "I like Postgres" and "I like SQL" are compared to "Rust is the best".  This is a artificial example, but the same idea holds true when translating to real models like `mixedbread-ai/mxbai-embed-large-v1`. 
```

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+-----|----------|
+| 1 | The pgml.transform function is a PostgreSQL function for calling LLMs in the database. |
+| 2 | I think tomatoes are incredible on burgers. |
+
+
+and a user is looking for the answer to the question: "What is the pgml.transform function?". If we embed the search query and all of the documents using a model like _mixedbread-ai/mxbai-embed-large-v1_, we can compare the query embedding to all of the document embeddings, and select the document that has the closest embedding in vector space, and therefore in meaning, to the to the answer. 
+
+These are big embeddings, so we can’t simply estimate which one is closest. So, how do we actually measure the similarity (distance) between different vectors?
+
+_pgvector_ as of this writing supports four different measurements of vector similarity:
```
montanalow, line 116:
```suggestion
`pgvector` as of this writing supports four different measurements of vector similarity:
```

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+- cosine distance
+- L1 distance
+
+For most use cases we recommend using the cosine distance as defined by the formula:
+
+<figure><img src=".gitbook/assets/cosine_similarity.png" alt="cosine similarity formula"><figcaption></figcaption></figure>
+
+where A and B are two vectors. 
+
+This is a somewhat confusing formula but luckily  _pgvector_ provides an operator that computes the cosine distance for us:
```
montanalow, line 129:
```suggestion
This is a somewhat confusing formula but luckily  `pgvector` provides an operator that computes the cosine distance for us:
```

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+These are big embeddings, so we can’t simply estimate which one is closest. So, how do we actually measure the similarity (distance) between different vectors?
+
+_pgvector_ as of this writing supports four different measurements of vector similarity:
+
+- L2 distance
+- (negative) inner product
+- cosine distance
+- L1 distance
+
+For most use cases we recommend using the cosine distance as defined by the formula:
```
montanalow, line 123: You may be able to shorten this section with a link to our docs on distance functions.

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+cosine_distance   
+--------------------
+ 0.7328613577628744
+```
+
+!!!
+
+!!!
+
+You'll notice that the distance between "What is the pgml.transform function?" and "The pgml.transform function is a PostgreSQL function for calling LLMs in the database." is much smaller than the cosine distance between "What is the pgml.transform function?" and "I think tomatoes are incredible on burgers".
```
montanalow, line 201: It's probably worth mentioning "out of sample" tokens, since they are rife on domain specific topics like this.

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+
+## Making it fast!
+
+It is inefficient to compute embeddings for all the documents every time we search the dataset. Instead, we should embed our documents once and search against precomputed embeddings. 
+
+_pgvector_ provides us with the `vector` data type for storing embeddings in regular PostgreSQL tables:
```
montanalow, line 207:
```suggestion
`pgvector` provides us with the `vector` data type for storing embeddings in regular PostgreSQL tables:
```

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+    md5(random()::text)
+  ) 
+FROM generate_series(1, 100000);
+```
+
+!!!
+
+!!!
+
+Now trying our search engine again:
```
montanalow, line 308: This would be better illustrated with a graph at 10, 100, 1000, 10000, 100k, 1M. Show that Big O.

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+                                          text                                          |   cosine_distance   
+----------------------------------------------------------------------------------------+---------------------
+ The pgml.transform function is a PostgreSQL function for calling LLMs in the database. | 0.13467974993681486
+```
+
+!!!
+
+!!!
+
+This somewhat less than ideal performance can be fixed by indexing the embedding column. There are two types of indexes available in _pgvector_: IVFFlat and HNSW.
```
montanalow, line 345:
```suggestion
This somewhat less than ideal performance can be fixed by indexing the embedding column. There are two types of indexes available in `pgvector`: IVFFlat and HNSW.
```

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+                                          text                                          |   cosine_distance   
+----------------------------------------------------------------------------------------+---------------------
+ The pgml.transform function is a PostgreSQL function for calling LLMs in the database. | 0.13467974993681486
+```
+
+!!!
+
+!!!
+
+We can see it is a massive speedup because we are only searching over 1/10th of the original vectors!
```
montanalow, line 400:
```suggestion
We can see it is a massive speedup because we are only comparing our input to 1/10th of the original vectors, instead of all of them! 
```

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+ The pgml.transform function is a PostgreSQL function for calling LLMs in the database. | 0.13467974993681486
+```
+
+!!!
+
+!!!
+
+We can see it is a massive speedup because we are only searching over 1/10th of the original vectors! 
+
+HNSW indexes are a bit more complicated. It is essentially a graph with edges linked by proximity in vector space. For more information you can check out this [write-up](https://www.pinecone.io/learn/series/faiss/hnsw/).
```
montanalow, line 402: Link to wikipedia if we have to link out, but links out usually end a session.

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,467 @@
...
+
+!!!
+
+!!!
+
+We can see it is a massive speedup because we are only searching over 1/10th of the original vectors! 
+
+HNSW indexes are a bit more complicated. It is essentially a graph with edges linked by proximity in vector space. For more information you can check out this [write-up](https://www.pinecone.io/learn/series/faiss/hnsw/). 
+
+HNSW indexes typically have better and faster recall but require more compute when adding new vectors. That being said, we recommend using HNSW indexes for most use cases.
```
montanalow, line 404:
```suggestion
HNSW indexes typically have better and faster recall but require more compute when adding new vectors. That being said, we recommend using HNSW indexes for most use cases where writes are less frequent than reads.
```

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
SilasMarvin: I'm not sure what that is?

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
SilasMarvin: I absolutely can, but do we want to? I was thinking this would post would give enough of an intro that after going through
it they would feel confident enough to reproduce this on their own test project, or rather, use this as a guide and copy the steps we did without too much hand waving. If the length of this section takes away from that goal I will trim it down!

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
montanalow: I agree that we could make some design improvements here. It'd be good to bring this up with the design team. We should still consistently use semantic markup, so that when the style is fixed, we get the benefits.

pgml-cms/blog/semantic-search-in-postgres-in-15-minutes.md
```diff
@@ -0,0 +1,494 @@
...
+
+!!!
+
+That was even faster!
+
+There is a lot more that can go into semantic search. Stay tuned for a follow up post on hybrid search and re-ranking.
+
+If you have any questions, or just have an idea on how to make PostgresML better, we'd love to hear from you in our [Discord](https://discord.com/invite/DmyJP3qJ7U). We’re open source, and welcome contributions from the community, especially when it comes to the rapidly evolving ML/AI landscape.
+
+## Closing thoughts / why PostgresQL?
```
levkk, line 484: ~PostgresQL~ PostgreSQL

PR #1539, opened 2024-06-19
Batch upsert documents

levkk:
### Features
1. [SDK] Add support for automatic batching for document upserts. Looking for review on the API before moving forward with PR, e.g. before adding tests.

```python
from pgml import Collection, Batch

collection = Collection("my_collection")
batch = Batch(collection, 25)

await batch.upsert_documents([{"id": 1}]) # Doesn't upsert yet

for i in range(23):
    await batch.upsert_documents([{"id": i}]) # Doesn't upsert yet

# Upserts whatever is in the current batch
# and appends the document to the next batch
await batch.upsert_documents([{"id": 1}])

# Upserts the final batch
await batch.finish()
```

### Bugs
1. [SDK] Formatting for long SQL queries.
3. [SDK] A couple of spelling typos.

### DIFF
--- a/pgml-sdks/pgml/Cargo.lock
+++ b/pgml-sdks/pgml/Cargo.lock
@@ -1590,7 +1590,7 @@ checksum = "e3148f5046208a5d56bcfc03053e3ca6334e51da8dfb19b6cdc8b306fae3283e"
 
 [[package]]
 name = "pgml"
-version = "1.0.4"
+version = "1.2.0"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/pgml-sdks/pgml/Cargo.toml
+++ b/pgml-sdks/pgml/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pgml"
-version = "1.1.0"
+version = "1.2.0"
 edition = "2021"
 authors = ["PosgresML <team@postgresml.org>"]
 homepage = "https://postgresml.org/"
--- a/pgml-sdks/pgml/src/batch.rs
+++ b/pgml-sdks/pgml/src/batch.rs
@@ -1,0 +1,101 @@
+//! Upsert documents in batches.
+
+#[cfg(feature = "rust_bridge")]
+use rust_bridge::{alias, alias_methods};
+
+use tracing::instrument;
+
+use crate::{types::Json, Collection};
+
+#[cfg(feature = "python")]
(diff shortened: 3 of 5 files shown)


SilasMarvin:
I'm not sure why a user would use that, and not just:
```python
from pgml import Collection, Batch

collection = Collection("my_collection")
# batch = Batch(collection, 25, {"merge": True})
batch = []

await batch.upsert_documents([{"id": 1}]) # Doesn't upsert yet
batch.append({"id": 1})

for i in range(23):
    # await batch.upsert_documents([{"id": i}]) # Doesn't upsert yet
    batch.append({"id": i})

# Upserts whatever is in the current batch
# and appends the document to the next batch
# await batch.upsert_documents([{"id": 1}])
await collection.upsert_documents(batch, {"merge": True})

# Upserts the final batch
# await batch.finish()
```

SilasMarvin:
Oh I see the automatic handling of upserting after they hit the threshold is nice, but it is a bit confusing. I think most people in the Python world are used to using batching systems already built into the dataset they are operating on. For example: https://huggingface.co/docs/datasets/en/process#batch-processing Not saying we shouldn't add it, but maybe we should clarify the name to like `AutoBatchUpsert` or something I'm not sure.

montanalow:
Why not use the `batch_size` argument on Collection.upsert_documents for this functionality?

levkk:
> Python world are used to using batching systems already built into the dataset

Datasets are one of very many sources for data. For example, my use case that triggered the desire for this feature was streaming WET files from a [`warcio.archiveiterator.ArchiveIterator`](https://github.com/webrecorder/warcio?tab=readme-ov-file#reading-warc-records) which seemingly doesn't have batching support built-in. Is it easy to write the batching logic yourself? Seemingly so, but it's really easy to forget to flush the last often incomplete batch when the source stream is complete.

> Why not use the `batch_size` argument on Collection.upsert_documents for this functionality?

`batch_size` doesn't handle the incomplete batch scenario, where a user inserts `len(records) % batch_size != 0`, hence the need for `finish()` aka [`flush()`](https://docs.rs/tokio/latest/tokio/io/struct.BufWriter.html#method.poll_flush).

montanalow:
Right, having to call flush/finish is only an issue because of this new API you’re introducing. The example Silas gave doesn’t have the issue.

PR #1555, opened 2024-07-12
Separate embedding kwargs into init kwargs and encode kwargs

### LINKED ISSUE #1169: pgml.embed trust_remote_code
A user on Discord submitted the following query with subsequent error:

```sql
SELECT pgml.embed(
    transformer => 'jinaai/jina-embeddings-v2-base-en',
    text => 'Dynamical Scalar Degree of Freedom in Horava-Lifshitz Gravity"}',
    kwargs => '{"trust_remote_code": true}'
);
```
```
ERROR:  ValueError: Loading this model requires you to execute custom code contained in the model repository on your localmachine. Please set the option trust_remote_code=True to permit loading of this model.
```

We are currently using `sentence_transformers` and `InstructorEmbedding` Python packages for the `pgml.embed` function. This model gets dispatched to the `sentence_transformers` package which does not appear to support the `trust_remote_code` keyword.

[Huggingface](https://huggingface.co/jinaai/jina-embeddings-v2-base-en#usage) shows, for this model, to use `transformers.AutoModel`. We should probably add this in, but how do we determine which model should use which package? Can they all use `transformers`? These questions need answering.

tomaarsen:
Resolves #1169

Hello!

## Pull Request overview
* Separate embedding kwargs into init kwargs and encode kwargs
* Introduces support for custom code models via `trust_remote_code` (e.g. #1169)
* Introduces support for private models via `token` (previously only possible via an environment variable, which FYI is still the recommended approach for security)
* Introduces support for Matryoshka models such as [this Vietnamese one](https://huggingface.co/hiieu/halong_embedding), which was trained such that embeddings can be truncated to smaller sizes with minimal performance loss & much faster retrieval, via `truncate_dim`.
* Introduces advanced loading support via `model_kwargs`/`tokenizer_kwargs`/`config_kwargs`. The first is most useful for inference, e.g. allowing loading models in lower precision for faster inference: `model_kwargs={"torch_dtype": "bfloat16"}`.

## Details
This PR splits `kwargs` in `pgml.embed` into two types of kwargs: for `model = SentenceTransformer(..., **kwargs)` and for `model.encode(..., **kwargs)`. This is currently done using a simple filter that checks for kwargs that are only (e.g. `trust_remote_code`) or primarily (e.g. `device`) relevant for the initialization.

I want to give a big preface that I have not tested this (!). My bandwidth is a bit too small this week for that I'm afraid. Another note is that `model_kwargs`/`tokenizer_kwargs`/`config_kwargs` and `truncate_dim` were only introduced in Sentence Transformers v3.0.0, whereas this project seems to be on v2.7 still. (FYI: ST v3.0 does not introduce breaking changes for inference, so upgrading should be safe).

- Tom Aarsen

### DIFF
--- a/pgml-extension/examples/embedding.sql
+++ b/pgml-extension/examples/embedding.sql
@@ -1,0 +1,7 @@
+\timing on
+
+SELECT pgml.embed('Alibaba-NLP/gte-base-en-v1.5', 'hi mom', '{"trust_remote_code": true}');
+SELECT pgml.embed('Alibaba-NLP/gte-base-en-v1.5', 'hi mom', '{"device": "cuda", "trust_remote_code": true}');
+SELECT pgml.embed('Alibaba-NLP/gte-base-en-v1.5', 'hi mom', '{"device": "cpu", "trust_remote_code": true}');
+SELECT pgml.embed('hkunlp/instructor-xl', 'hi mom', '{"instruction": "Encode it with love"}');
+SELECT pgml.embed('mixedbread-ai/mxbai-embed-large-v1', 'test', '{"prompt": "test prompt: "}');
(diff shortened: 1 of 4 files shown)


montanalow:
Thanks for the PR. I've added our embedding tests to CI, since we generally don't run the whole transformers suite due to the model download times. Confirmed that `trust_remote_code` flag now works as expected.

tomaarsen:
Excellent, thank you for merging & writing some simple tests.

- Tom Aarsen

PR #1568, opened 2024-07-16
Update README.md

cassandrastumer:


montanalow (changes requested):
Probably shouldn't delete words

montanalow (approved):

PR #1586, opened 2024-07-23
llama 3.1 support

montanalow:


pgml-extension/requirements.linux.txt
```diff
@@ -89,85 +85,85 @@ nvidia-cufft-cu12==11.0.2.54
...
 tokenizers==0.19.1
-torch==2.3.0
-torchaudio==2.3.0
-torchvision==0.18.0
+torch==2.3.1
+torchaudio==2.3.1
+torchvision==0.18.1
 tqdm==4.66.4
-transformers==4.40.2
+transformers==4.42.4
```
kczimm, line 151: We want `transformers==4.43.1` for the new Llama-3.1 models. See [here](https://huggingface.co/meta-llama/Meta-Llama-3.1-8B-Instruct#use-with-transformers).

PR #1636, opened 2024-10-11
Fix bug that shape mismatch error in predict when changing objective to softmax and update rust-xgboost commit

cyccbxhl:
Fix bugs about rust-xgboost:
        1) NaN recall;
        2) Shape mismatch error in predict when changing objective to softmax.


montanalow:
It would be nice if XGBooster had a get_param, so we didn't need to pass hyperparams around like this, but it doesn't, so we do. Thanks for fixing this up.

